PR #32, opened 2018-05-31
add default pre-commit git hook

jalimena:
You can set up this git hook to run whenever you try to commit something. It is the default available one on git. It basically just checks that there is no trailing whitespace in your commit.


bpf6qc:
This seems like a smart change that we should have everyone use. I would additionally propose that we add the commands to enable this to changeDataFormats.py so that it's done automatically.

And on that thought since changeDataFormats.py is now mandatory to get the CMSSW version set, perhaps we should change its name to be more evocative of its purpose. Maybe "setup.py" or "setupFramework.py" or something along those lines?

jalimena:
Hi @bpf6qc , I'm a bit confused. Why do you say we should add this to changeDataFormats.py? The point is that for each new commit, you check for trailing whitespace and the like.

If changeDataFormats.py is still only changing the data format, then maybe we keep the name as it is, as that is more descriptive...

bpf6qc:
Your ELOG post said that if users wanted to enable this check, they need to make a link in the git structure somewhere. I'm suggesting that this not be made a choice, and that this be done for the user automatically. The way to do that would be, if there was a script that was mandatory to run once when setting up a release, we could add this link-creation to that script. changeDataFormats.py is such a script.

And yes changeDataFormats.py does chiefly change the data format, but we had the same requirement of detecting the CMSSW release once when setting up the area -- otherwise the "numberOfHits" --> "numberOfAllHits" would never compile. So in the same case we needed a script that people would always run once when setting it up, and this CMSSW_VERSION issue was handled there.

jalimena:
thanks for explaining, Bryan. new commits should do as you suggested

jalimena:
superseded by #33

PR #34, opened 2018-06-05
Electron ID overhaul

bpf6qc:
- Introduce VID (versioned identification) methods for cut-based electron IDs. WPs for 2015, 2016, and 2017.
- Remove electron effective area magic numbers in favor of POG payload files.
- Introduce several new methods and update the magic number (old) version of the electron tight ID. This method remains for comparison and backwards compatibility.
- Add HLT_Ele35_WPTight_Gsf_v to hlt-matching.
- Add photon effective areas -- this is actually an ancient change in my forked repo that can be added as well.

The electron cut-based VID IDs are available by default when the user sets up their working area to use MINIAOD. The working points are chosen automatically depending on which release they are in. Specifically:

76X: `cutBasedElectronID-Spring15-25ns-V1-standalone-*`
80X: `cutBasedElectronID-Summer16-80X-V1-*`
94X: `cutBasedElectronID-Fall17-94X-V1-*`

Many other VID IDs are available such as MVA and HEEP working points, which can be used either by changing the tags in `collectionProducer.electrons.vid*IdMap` or by introducing new functions similar to `osu::Electron::passesVID_*ID` that uses these.

When using 80X releases, the following commands are REQUIRED in your recipe:
```
git cms-merge-topic ikrav:egm_id_80X_v2
cd $CMSSW_BASE/external/$SCRAM_ARCH/
git clone https://github.com/lsoffi/RecoEgamma-ElectronIdentification.git data/RecoEgamma/ElectronIdentification/data
cd data/RecoEgamma/ElectronIdentification/data
git checkout CMSSW_9_4_0_pre3_TnP
cd $CMSSW_BASE/src
```
Unfortunately the default 80X code makes reference to MVA weights that are not available, and this reference is made regardless of whether the MVA IDs are actually being produced.

These VIDs are run using the entire electrons collection currently. In the future the input collections could be modified intelligently within our framework such that different VID IDs could be recalculated on different stages of electron selection, but this is not now supported.

In the case that a user needs to remove a cut from the VID ID, most notably the isolation requirement, the user can use:
`removeVIDCut(process, 'cutBasedElectronID-Summer16-80X-V1-tight', 'GsfEleEffAreaPFIsoCut')`
or similar. This should be considered an expert procedure as it defeats the purpose of the VID. The user should consider whether to use this or to define their non-isolated ID by hand in python cuts. This function should be called at the end of your protoConfig file, whatever file is the target of the `cmsRun` command, to be certain that all instances of this cut are modified as desired.

The VID tight ID has been observed to be identical to the old "passesTightID_noIsolation" method under the following circumstances (and more below in replies):

- 10k events from root://cmsxrootd.fnal.gov//store/mc/RunIIFall17MiniAODv2/DYJetsToLL_M-50_TuneCP5_13TeV-madgraphMLM-pythia8/MINIAODSIM/PU2017RECOSIMstep_12Apr2018_94X_mc2017_realistic_v14-v1/20000/F22CA5A8-7D44-E811-8509-44A842CFD667.root
- CMSSW 9_4_8

```
JustTightIDTest channel:
----------------------------------------------------------------------------------------------
Cut Name                                                Events     Cumul. Eff.     Indiv. Eff.
----------------------------------------------------------------------------------------------
total                                                   9996.0        100.000%        100.000%
>= 1 electrons with pt > 5                              4426.0         44.278%         44.278%
>= 1 electrons with fabs(superCluster.eta) < 2.5        4342.0         43.437%         56.523%
>= 1 electrons with passesTightID_noIsolation > 0           2390.0         23.910%         23.910%
>= 1 electrons with tight #rho-corrected rel. iso.      2127.0         21.279%         49.600%
----------------------------------------------------------------------------------------------
with tight #rho-corrected rel. iso. = ((fabs (superCluster.eta) <= 1.479) && (((pfIso_.sumChargedHadronPt + max (0.0, pfIso_.sumNeutralHadronEt + pfIso_.sumPhotonEt - rho * AEff)) / pt) < 0.0361)) \
                         || ((fabs (superCluster.eta) >  1.479) && (((pfIso_.sumChargedHadronPt + max (0.0, pfIso_.sumNeutralHadronEt + pfIso_.sumPhotonEt - rho * AEff)) / pt) < 0.094))

JustVIDTest channel:
------------------------------------------------------------------------------------------------
Cut Name                                                  Events     Cumul. Eff.     Indiv. Eff.
------------------------------------------------------------------------------------------------
total                                                     9996.0        100.000%        100.000%
>= 1 electrons with passesVID_tightID > 0                 2338.0         23.389%         23.389%
>= 1 electrons with |d0| < 0.05:0.10 (barrel:endcap)      2337.0         23.379%         52.261%
>= 1 electrons with |dz| < 0.10:0.20 (barrel:endcap)	  2127.0         21.279%         36.595%
------------------------------------------------------------------------------------------------
where |d0| = ((fabs (electron.superCluster.eta) <= 1.479) && (fabs(((electron.vx - eventvariable.leadingPV_x) * electron.py - (electron.vy - eventvariable.leadingPV_y) * electron.px) / electron.pt) < 0.05)) \
                        ||  ((fabs (electron.superCluster.eta) >  1.479) && (fabs(((electron.vx - eventvariable.leadingPV_x) * electron.py - (electron.vy - eventvariable.leadingPV_y) * electron.px) / electron.pt) < 0.10))
and |dz| = ((fabs (electron.superCluster.eta) <= 1.479) && (fabs (electron.vz - eventvariable.leadingPV_z) < 0.10)) \
                        ||  ((fabs (electron.superCluster.eta) >  1.479) && (fabs (electron.vz - eventvariable.leadingPV_z) < 0.20))
```

So VID tight ID includes pt>5, |eta|<2.5, tight ID, tight PF rel iso, but no vertexing.

VID tags are controlled with collectionProducer.electrons.vidTightIdMap (et al) in OSUT3Analysis/Configuration/python/CollectionProducer_cff.py and are keyed off of CMSSW release. MVA and HEEP electron IDs are easily added in a similar fashion with their relevant tags.

One important note: numberOfAllHits counts the number of hits in a range of hits, but numberOfLostHits counts the MISSING hits in a range of hits. The difference being that AllHits would count +1 for an INACTIVE hit, but LostHits do not count such hits. Notable for disappearing tracks but for electrons this is the way the POG and VID does it.


Collections/src/Electron.cc
```diff
@@ -235,38 +272,105 @@ void
...
-                                 && fabs (this->gsfTrack ()->dxy (vertex.position ()))                                    <   0.0351
-                                 && fabs (this->gsfTrack ()->dz (vertex.position ()))                                     <   0.417
+
 #if CMSSW_VERSION_CODE >= CMSSW_VERSION(9,4,0)
-                                 && this->gsfTrack ()->hitPattern ().numberOfAllHits (reco::HitPattern::MISSING_INNER_HITS)  <=  1
-#else
-                                 && this->gsfTrack ()->hitPattern ().numberOfHits (reco::HitPattern::MISSING_INNER_HITS)  <=  1
-#endif
-                                 && !ConversionTools::hasMatchedConversion (*this, conversions, beamspot.position ()));
+  // https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedElectronIdentificationRun2#Offline_selection_criteria
```
jalimena: I don't understand several things this section:
- Why should we keep anything in Electron.cc that explicitly references all the cuts in a given ID? Isn't the point to remove all things like this? In other words, why define passesTightID_noIsolation_ at all?
- How does it make sense without the &&s?
- Why does this definition of passesTightID_noIsolation_ include dxy and dz cuts, when the electron ID twiki says that they are not included in the ID any more?
- I understand that numberOfLostHits is better for disappearing tracks, but numberOfAllHits is probably what should be used for displaced SUSY. So, what in the end is implemented? I'm confused.

bpf6qc:
"Why should we leave it up to the user to choose": these are good points, the immediate answer is that I didn't find a place to edit the cmsRun process to insert the producers. Now that you ask these, this function can probably be called in add_channels with some minor checks. I'll figure out how to implement this.

The tags need to be easily configurable however, because in the case of disappearing tracks for example we used the 2015 ID for both 2015 and 2016 data since it was looser (ie a tighter veto) and for consistency. The method to change these can be obscured, but it should still be available.

"shouldn't we implement something similar for": yes we should. In some cases like muon ID we have it easy because CMSSW provides these methods. But in general we should review the existing code and search for these magic numbers and the way to remove them.

jalimena:
Thanks, @bpf6qc . All makes sense. I've made an issue ( #35  ) so that we don't forget to do this.

Collections/src/Electron.cc
bpf6qc: - I kept this for the moment for backwards compatibility and to compare the two methods. I agree that the point is to remove it and we should, but I'd like to review this in light of disappearing tracks first. Specifically, since the isolation is baked into the VID, if we need this WITHOUT isolation then I need to find a method to accomplish that.
- The &&s are just moved to the right side of the lines. This is purely a conventional change, I just don't think any other such lines are done with the &&s on the left in this repo.
- This is probably an artifact of Andrew writing it for disappearing tracks and Jamie never using this function for displaced SUSY. Disappearing tracks requires these cuts, and no one complained that they were in it.
- It's not necessarily better for disappearing tracks, I just meant that it's an interesting and easily overlooked change. The VID in 94X uses LostHits, which is outside of our control, and I changed our own function to match EGM's logic. In the end LostHits is a looser cut than AllHits here -- in my comparison test I saw one electron that had a hit pattern of (INVALID, MISSING) which failed AllHits<=1 but passed LostHits<=1 and passed the VID. For displaced SUSY, you will need to ask yourself if such cases are acceptable for your analysis, but we should just highlight that such a thing is happening when the VID is used.

bpf6qc:
Please wait to merge this request until I've reviewed these changes in 76X and 80X. Also need to look for cases when isolation needs to be removed and how to do this, as well as change the customize function to be default.

btcardwell:
Thanks for implementing this Brian. I totally support the general goal here and think your answers to Juliette's questions above make sense. I'll look through the commits in more detail and speak up if I have questions or concerns.

bpf6qc:
Under the same conditions as the previous test (same file, events, 94X), a call to:
```removeVIDCut(process, 'cutBasedElectronID-Fall17-94X-V1-tight', 'GsfEleEffAreaPFIsoCut')```

is shown to work as intended by comparing:
```
JustVIDTest channel:
------------------------------------------------------------------------------------------------
Cut Name                                                  Events     Cumul. Eff.     Indiv. Eff.
------------------------------------------------------------------------------------------------
total                                                     9996.0        100.000%        100.000%
>= 1 electrons with passesVID_tightID > 0                 2662.0         26.631%         26.631%
>= 1 electrons with |d0| < 0.05:0.10 (barrel:endcap)      2661.0         26.621%         52.261%
>= 1 electrons with |dz| < 0.10:0.20 (barrel:endcap)	  2390.0         23.910%         36.595%
------------------------------------------------------------------------------------------------

JustTightIDTest channel:
----------------------------------------------------------------------------------------------
Cut Name                                                Events     Cumul. Eff.     Indiv. Eff.
----------------------------------------------------------------------------------------------
total                                                   9996.0        100.000%        100.000%
>= 1 electrons with pt > 5                              4426.0         44.278%         44.278%
>= 1 electrons with fabs(superCluster.eta) < 2.5        4342.0         43.437%         56.523%
>= 1 electrons with isTightElectronWRTVtx > 0           2390.0         23.910%         23.910%
>= 1 electrons with tight #rho-corrected rel. iso.      2127.0         21.279%         49.600%
----------------------------------------------------------------------------------------------
```

**Comment by jalimena** [comment]
Hi @bpf6qc , thanks for the latest commit. Is there anything else on this PR that you are working on? Or would you consider it ready to be merged?

**Comment by bpf6qc** [comment]
I'm currently testing this in 76X and 80X, I'd like to run these two channels (above) and make sure it's not broken.

**Comment by jalimena** [comment]
Ok thanks. I'll wait for your report back here then. 

Also, I would ask that you modify the top of this PR with:
- how the user should use this, like if the process = customizeMINIAODElectronVID (process, "Fall17") statement has now changed
- if anything else has changed, any other important things to note

Thanks!

**Comment by bpf6qc** [comment]
Comparison 76X over 1k events in /store/user/ahart/WJetsToLNu_HT-600ToInf_TuneCUETP8M1_13TeV-madgraphMLM-pythia8/RunIIFall15MiniAODv2-PU25nsData2015v1_76X_mcRun2_asymptotic_v12-DisappTrks-v1/160205_142511/0000/miniAODWithCandidateTracks_1.root

```
JustVIDTest channel:
----------------------------------------------------------------------------------------------------
Cut Name                                                      Events     Cumul. Eff.     Indiv. Eff.
----------------------------------------------------------------------------------------------------
total                                                         1000.0        100.000%        100.000%
>= 1 electrons with passesVID_tightID > 0                      220.0         22.000%         22.000%
>= 1 electrons with |d0| < 0.0111:0.0351 (barrel:endcap)       220.0         22.000%         62.100%
>= 1 electrons with |dz| < 0.0466:0.417 (barrel:endcap)        220.0         22.000%         65.600%
----------------------------------------------------------------------------------------------------

JustTightIDTest channel:
----------------------------------------------------------------------------------------------
Cut Name                                                Events     Cumul. Eff.     Indiv. Eff.
----------------------------------------------------------------------------------------------
total                                                   1000.0        100.000%        100.000%
>= 1 electrons with pt > 5                               642.0         64.200%         64.200%
>= 1 electrons with fabs(superCluster.eta) < 2.5         639.0         63.900%         73.000%
>= 1 electrons with isTightElectronWRTVtx > 0            220.0         22.000%         22.000%
>= 1 electrons with tight #rho-corrected rel. iso.       175.0         17.500%         44.300%
----------------------------------------------------------------------------------------------
```

**Comment by bpf6qc** [comment]
Comparison in 80X fails because passesTightID_noIsolation is intentionally an older 2015 recommendation.

Further 80X will only work with the additional recipe steps:
```
git cms-merge-topic ikrav:egm_id_80X_v2
cd $CMSSW_BASE/external/$SCRAM_ARCH/
git clone https://github.com/lsoffi/RecoEgamma-ElectronIdentification.git data/RecoEgamma/ElectronIdentification/data
cd data/RecoEgamma/ElectronIdentification/data
git checkout CMSSW_9_4_0_pre3_TnP
cd $CMSSW_BASE/src
```
This is unfortunately necessary when using any VID at all in 80X as the default code references MVA weights not available by default, no matter if the user actually requests the MVA ID or not.

bpf6qc:
I've updated the main PR description above. The final issue remaining here, for me, is the question of removing passesTightID_noIsolation entirely. Disappearing tracks never inverts the isolation requirement on electrons for any channels, so this is not a problem for us.

jalimena:
ok thanks, @bpf6qc . I think we can merge this now, and remove passesTightID_noIsolation later on. Do you agree? If so, I'll merge this (or someone else can).

bpf6qc:
@jalimena that sounds good for me, I agree.

jalimena:
+1

PR #39, opened 2018-06-13
Skim fix for ele VIDs

bpf6qc:
c.f. https://cmshead.mps.ohio-state.edu:8080/OSUT3Analysis/117

For skims:
Saves CaloCluster collections so that electron->superCluster()->seed() can be called.
Change the input electron collection to what's stored in the skim for the electron VID calculation.


jalimena:
this PR seems in principle fine. the only question is when to merge and if we should remake the skims.

bpf6qc:
The issue with changing the input tag for electrons when re-running the VIDs, that would just prevent you from running over a skim and can be added later.

The important issue for skims is that we only just started using `electron->superCluster()->seed() `last week, and every skim we've ever made so far lacks the CaloCluster collection to make this work. We could just use `dEtaIn` instead of `dEtaInSeed` but then we're not following EGM recommendations. Also the VIDs inherently use `seed()` now, so we would need to turn the VIDs off entirely.

One option we have to save our existing skims is giving the original MINIAOD as secondaryFiles. That should fetch the missing CaloClusters for `seed()` and work just fine. We've done this before in individual cases, but I'm looking into making this a general case now.

PR #40, opened 2018-06-19
Save all content in skims

bpf6qc:
c.f. https://cmshead.mps.ohio-state.edu:8080/OSUT3Analysis/118

### DIFF
--- a/Configuration/python/mergeUtilities.py
+++ b/Configuration/python/mergeUtilities.py
@@ -175,7 +175,7 @@ def MakeFilesForSkimDirectory(Directory, DirectoryOut, TotalNumber, SkimNumber, 
 ###############################################################################
 #           Produce a pickle file containing the skim input tags.             #
 ###############################################################################
-def GetSkimInputTags(File, ignoreSkimmedCollections=False):
+def GetSkimInputTags(File):
     print "Getting skim input tags..."
     eventContent = subprocess.check_output (["edmDumpEventContent", "--all", os.getcwd () + "/" + File])
     parsing = False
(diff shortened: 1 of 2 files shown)


bpf6qc:
@jalimena if there are any custom keep commands for DisplacedSUSY or StoppedPtls, such as:
https://github.com/OSU-CMS/DisappTrks/blob/master/CandidateTrackProducer/python/customize.py#L3-L12
then these should be added to the outputCommands as well, to save those originals.

jalimena:
hi @bpf6qc 
For DisplacedSUSY, there are no custom keep commands, we just simply use the MINIAOD event content.
For Stopped Particles, though, it's more complicated. We keep and drop different things for the jets channel versus the muon channel. Take a look at:
https://github.com/allenji/StoppPtls/blob/StpPtls_9XX/Collection/test/delayedMuonsCandiate_MC_cfg.py#L76-L82
https://github.com/allenji/StoppPtls/blob/StpPtls_9XX/Collection/python/stoppPtlsCandidate_MC_2016.py#L89-L98
https://github.com/allenji/StoppPtls/blob/StpPtls_9XX/Collection/python/stoppPtlsCandidate_cfi.py
https://github.com/allenji/StoppPtls/blob/StpPtls_9XX/Collection/python/stoppPtlsJetsCandidate_cfi.py
So I'm not quite sure how to proceed to keep everything ok for Stopped Particles (to jets and to muons).

bpf6qc:
@jalimena If the StpPtls repo were to change those config files to define a vstring object and then extend the outputCommands, like in:
https://github.com/OSU-CMS/DisappTrks/blob/master/CandidateTrackProducer/python/customize.py#L3-L12
then we could import that object in the same way I've done with that. Otherwise it's not dangerous to append more keep commands, we can just append more commands.

jalimena:
hi @bpf6qc , sorry for the delay.
I've done as you suggested, please see https://github.com/allenji/StoppPtls/blob/StpPtls_9XX/Collection/python/outputCommands.py
I think with this, we should be set.

jalimena:
In principle, this all seems fine to me. However, I would prefer if @aehart takes a look and comments. @bpf6qc , from your side, do you think this PR is complete now? You've tested it?

If we are ready to merge, @bpf6qc , can you make an elog with this change and make it clear what the user should do?

PR #43, opened 2018-06-29
Fix osusub.py for UserDir and UserList modes

bpf6qc:
Write an empty listOfSecondaryFiles for these modes. They are used in
signal MC generation where there are no secondary files.

PR try 2 after a rebase.


jalimena:
@aehart  can you please review this PR?

PR #45, opened 2018-07-06
Add deltaRToClosest(Lepton) methods to osu::Track

bpf6qc:
Only defined if DISAPP_TRKS is defined by setupFramework.py.


zacharypollock:
still debugging. i'm having some issues and it's still unclear whether they are personal or with the PR. 

Please refrain from merging until a new comment is made here

bpf6qc:
Zach encountered an error that was later shown to be a personal one. I've also made a small change after comments from Andrew.

PR #46, opened 2018-07-09
add TTJets inclusive sample for 94X and QCD DoubleEMEnriched samples

jalimena:


jalimena:
I think anyone could review and merge this PR, it should be straightforward and not controversial. Thanks!

PR #50, opened 2018-07-17
add 2017 MuEG and DoubleMu composite datasets without 2017B to configurationOptions

jalimena:
Since the Displaced Lepton muon-photon and mu-mu triggers were not available throughout 2017B, adding composite dataset definitions in which all of 2017 data is used except 2017B. Done for the MuonEG and DoubleMuon datasets. Should be a straightforward PR.


aehart:
This is a trivial change, which I'll go ahead and approve and merge.

PR #53, opened 2018-07-18
ele VID fixes for DisappTrk track methods

bpf6qc:
Remove MVA and HEEP ID VIDs for electrons
Fix specific case for DisappTrks with electrons + tracks
Require vertexing for DisappTrks track::deltaRToClosest*Electron methods

### DIFF
--- a/Collections/src/Track.cc
+++ b/Collections/src/Track.cc
@@ -331,7 +331,7 @@ osu::Track::set_minDeltaRToElectrons (const edm::Handle<edm::View<TYPE(electrons
     // Note in below, these remain false if |eta| >= 2.5; thus an eta cut is also being applied here as intended
 #if CMSSW_VERSION_CODE < CMSSW_VERSION(9,4,0)
     // https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedElectronIdentificationRun2Archive#Spring15_selection_25ns
-    if(fabs(electron.superCluster ()->eta()) <= 1.479) {
+    if(fabs(ele.superCluster ()->eta()) <= 1.479) {
       passesVeto_dxy = (fabs(ele.gsfTrack()->dxy(vertices->at(0).position())) < 0.0564);
       passesVeto_dz  = (fabs(ele.gsfTrack()->dz(vertices->at(0).position()))  < 0.472);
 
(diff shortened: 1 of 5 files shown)


Collections/src/Track.cc
```diff
@@ -321,11 +322,81 @@ osu::Track::set_minDeltaRToElectrons (const edm::Handle<edm::View<TYPE(electrons
...
-    if((*vidTightMap) [(*electrons).refAt(iEle)] && (dR < deltaRToClosestTightElectron_  || deltaRToClosestTightElectron_  < 0.0)) deltaRToClosestTightElectron_  = dR;
-  }
+
+    bool passesVeto_dxy   = false, passesVeto_dz   = false;
+    bool passesLoose_dxy  = false, passesLoose_dz  = false;
+    bool passesMedium_dxy = false, passesMedium_dz = false;
+    bool passesTight_dxy  = false, passesTight_dz  = false;
+
+    // Note in below, these remain false if |eta| >= 2.5; thus an eta cut is also being applied here as intended
+#if CMSSW_VERSION_CODE < CMSSW_VERSION(9,4,0)
```
jalimena: L332-370: Why are all these dxy cut values hardcoded?

bpf6qc: The EGM POG doesn't include vertex requirements in the VIDs for various reasons, so these cuts need to be made explicitly (they were forgotten before this PR by mistake). Plus this isn't the normal framework "electrons" collection, so they can't be provided in the typical cut PSet fashion.

Are you suggesting however that these be external PSet parameters?

jalimena: Ah ok, thanks for the information. But yeah, wouldn't external PSet parameters be a good idea?

bpf6qc: I can do that; it'll look cleaner too.

jalimena:
hi @bpf6qc , I'm going to run a test with this PR now to check that it doesn't break anything for displaced leptons. But also, what's the status of doing the modification we talked about above?

bpf6qc:
@jalimena I haven't had the time to do that yet, but I intend to within the week.

bpf6qc:
Testing recent commit...

bpf6qc:
@jalimena I'm satisfied with my changes now. These methods are all behind
```#ifdef DISAPP_TRKS```
so the potential to affect other analyses should be zero by design.

jalimena:
cool, looks good

PR #58, opened 2018-07-27
Handle missing dataset directories in mergeOutputHadd.py

btcardwell:
Don't try to merge datasets that don't exist in the condor directory; skip them and warn the user instead.


btcardwell:
Works as expected when tested on /data/users/bcardwell/condor/Pileup_18_03_12/.

btcardwell:
Ah, my bad on letting the merge commit slip in here. I'm not quite sure how that happened.

btcardwell:
Rebase success! Good thinking

PR #59, opened 2018-07-27
getTrigObj reject multiple strings

bpf6qc:
Reject multiple filter label substrings in `getTriggerObjectsByFilterSubstring` by providing a vector of substrings instead of just one.

Also clean up this function for readability.


jalimena:
this seems fine to me. if you have tested it, i think we can merge it.

PR #60, opened 2018-07-30
Fix empty skim issue

bpf6qc:
Fixes: empty skim outputs were only produced if the input file started
with “file:”, thus never if it was “root:”.

Introduces flag “forceNonEmptySkim” in case a user really wants a
non-empty skim.

Add “drop osu*_*_originalFormat_*” to never save transient internal osu
collections.

### DIFF
--- a/Configuration/python/processingUtilities.py
+++ b/Configuration/python/processingUtilities.py
@@ -502,14 +502,6 @@ def add_channels (process,
             else:
                 outputCommand += "*"
             outputCommands.append (outputCommand)
-
-        ########################################################################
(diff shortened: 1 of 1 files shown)


bpf6qc:
Tested the following:
1) On the T3 run over a skim_0.root file with XrootD (cmseos.fnal.gov) -- emptySkim.root correctly produced.
2) As 1 but setting forceNonEmptySkim = True correctly yields a non-empty skim.root. Furthermore the skim.root file does not contain any vector<osu::*> collections.
3) Running over either skim.root or emptySkim.root (1 or 2 above), the ambiguous product error is not encountered.

PR #62, opened 2018-07-31
Warn user about datasets with no input skim files instead of crashing

btcardwell:
Calling SkimModifier before copying datasetInfo lead to an error when running over a dataset with no skim files. This implementation warns the user and skips the dataset instead of giving an error and halting the submission process.

### DIFF
--- a/DBTools/scripts/osusub.py
+++ b/DBTools/scripts/osusub.py
@@ -381,6 +381,18 @@ def GetListOfRemoteRootFiles(Path):
                 fileList.append(filePath.split()[4])
     return fileList
 
+def CheckIfRemoteDirectoryExists(Path):
+    if not len(Path.split('//')) == 3 or not Path.startswith('root://'):
+        print 'Remote directories must be of the form \"root://<redirector>//store/...\"'
+        return False
+    thisRedirector = Path.split('//')[1] 
+    thisRemotePath = '/' + Path.split('//')[2]
+    try:
+        subprocess.check_call('xrd '+ thisRedirector + ' existdir ' + thisRemotePath, shell = True)
+        return True
+    except subprocess.CalledProcessError:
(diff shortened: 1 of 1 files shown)


btcardwell:
Works as expected when tested at the LPC. Test results here: /uscms/home/cardwell/nobackup/condor/EEPromptControlRegion_2016Analysis_30July2018

jalimena:
This is ok, but it's not yet implemented for when you have a remote skim. I'm working on it now.

jalimena:
I think I've at least started to implement it now, with my commit, for remote skims. I think it needs more testing, I'm not sure it's complete.

jalimena:
Can anyone else take a look at this PR, particularly for when we want to run over remote skims (but there's no input skim file)? When I submit remote skim jobs, I often get "[!] Some of the requests failed. The result may be incomplete." but I'm not sure if this is because some input skims aren't there or because of something else. Furthermore, most of my jobs are being held now. Probably I need help - if you can check my commit and fix any problems, that would be good. @btcardwell or @bpf6qc , for example?

bpf6qc:
I always get `[!] Some of the requests failed. The result may be incomplete.` while using `xrdfs ls -l -u` even when I check that the results are indeed complete, so I'm not sure how we can prevent possible errors.

Recent commit cleans a few things up; checking for a directory is done in the process of trying to list it, rather than checking multiple times. Also checks for correct copying of associated files.

Also fixed a small issue where the `originalNumberOfEvents` and `skimNumberOfEvents` would be erroneously set to zero in this mode.

I have successfully run one job on the OSU T3, with an input file on LPC EOS. I have also successfully seen an error message and not a crash for remote directories that do not exist.

btcardwell:
Looks good to me!

PR #65, opened 2018-08-09
Add several 2018ABC datasets

bpf6qc:


Configuration/python/configurationOptions.py
```diff
@@ -2603,6 +2651,9 @@
     'MET_2017D' : -1,
     'MET_2017E' : -1,
     'MET_2017F' : -1,
+    'MET_2018A' : -1,
```
jalimena: You've defined maxEvents for SingleMu_2018 and SingleEle_2018, but not for MET_2018 or Tau_2018. Maybe you want to do that?

Configuration/python/configurationOptions.py
```diff
@@ -4229,6 +4303,9 @@
     'SingleEle_2017D'   : "data",
     'SingleEle_2017E'   : "data",
     'SingleEle_2017F'   : "data",
+    'SingleEle_2018A'   : "data",
```
jalimena: You haven't defined SingleEle_2018 as data.

jalimena:
Also, I believe 2018 data is being reconstructed for now in 10_1_X (and 10_2_X in Run2018D), so the next step will be making sure the framework works in these releases.

PR #67, opened 2018-08-13
New track isolation types

zacharypollock:
This splits up Track.h into two objects: TrackBase.h (for general use) and DisappearingTrack.h (fo DisappTrk use).

Two notes:
1) When trying to test this with a DisplacedSUSY analysis to make sure we didn't break anything, EE and MuMu channels' preselection ran fine, but EMu gave me a "ProductNotFound" error  
error . 
This held true when I switched and tried the same thing with the clean master of OSUT3 and DisplacedSUSY repositories, so I am inclined to think this is not an issue that this change introduced.

2) @DisappTrk users: when this is approved, we need to remember to change two files in the DisappTrks repository to include DisappearingTrack.h instead of Track.h . The two files are:
    -- BackgroundEstimation/plugins/EventTPProducer.h
    -- StandardAnalysis/plugins/TriggerWeightProducer.cc


btcardwell:
The issue with DisplacedSUSY is just due to a miscommunication on my part: I forgot to tell Zach to specify in preselection_cfg.py that he was running over data before trying cmsRun.

I tested this PR in DisplacedSUSY and see no problems.

PR #68, opened 2018-08-16
Produce error if PUScaleFactorProducer denominator histogram is empty

btcardwell:
Tested interactively with empty and non-empty MC PU histograms. Behavior is unchanged with non-empty histograms and cmsRun exits with expected error for empty histograms.


bpf6qc:
Maybe here `mc->GetEntries() == 0` would be more appropriate, to avoid float precision comparisons and the never-ever-case of negative PU bins?

btcardwell:
Good thinking Brian. I almost asked if anyone had an idea for avoiding the floating point comparison. The most recent commit implements your suggestion and performs identically when tested.

PR #69, opened 2018-08-16
Create category for jobs removed from condor

bpf6qc:
At the LPC jobs can fail with “PERIODIC_SYSTEM_REMOVE” due to over-usage of memory or some other job parameter, and this would show up as “Running”.

A new category is created to make this clear to users.

### DIFF
--- a/AnaTools/scripts/checkJobs
+++ b/AnaTools/scripts/checkJobs
@@ -29,18 +29,24 @@ then
       PASS_ARR=($PASS_LOGS)
       PASS=${#PASS_ARR[@]}
     fi
+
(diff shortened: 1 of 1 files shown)


bpf6qc:
Currently it's possible for a job to appear both removed and successful. I need to fix this and I want to add some print options, so please wait.

bpf6qc:
For jobs that were removed, resubmitted, and are again running as checkJobs is called, they are now correctly considered as running.

Also now accepts several flags to print job ranges explicitly for each category, so `checkJobs condor/workDirectory --printFail` will print exactly which jobs are amongst the failed.

PR #71, opened 2018-08-23
Produce d0 smearing value in OSUElectronProducer

btcardwell:
This PR isn't ready to be merged, but I'd be interested in feedback if anyone has time.

I'm looking to smear the d0 values of electrons and muons by randomly sampling a Gaussian of a specified width and adding the result to the original d0 value. This PR currently only deals with electrons; I'll mimic the implementation in muons once everything comes together.

I initially thought to do this in DisplacedSUSYEventVariableProducer, but I was unable to figure out how to store d0 smearing values for each of the arbitrary number of electrons or muons in each event. I then switched to producing the d0 smearing value for each electron in OSUElectronProducer and adding it to the initial d0 when making histograms. This basic approach works, but there are a couple necessary features that I'm not sure how to implement. I'll keep working on it, but I figured I'd ask in case anyone has a quick answer.

Specifically, I have the following questions:
- In OSUElectronProducer, how can I check if an electron is from data or MC?
- Can I read in a value from a config file so I don't have to hard code magic numbers? In DisplacedSUSYEventVariableProducer I would use `cfg.getParameter`, but doing this here would force all analyses to include something like `d0SmearingWidth = 0.0` in their config files. Is there a better way?

### DIFF
--- a/Collections/interface/Electron.h
+++ b/Collections/interface/Electron.h
@@ -24,7 +24,8 @@ namespace osu
         const double sumChargedHadronPtCorr () const;
         const double sumPUPtCorr () const;
         const int electronPVIndex () const;
-	const double genD0 () const;
+        const double genD0 () const;
+        const double d0SmearingVal () const;
         const float dEtaInSeed () const;
         const bool pass_GsfEleHadronicOverEMEnergyScaledCut (const float c0, const float cE, const float cR) const;
 
@@ -52,6 +53,7 @@ namespace osu
         void set_sumPUPtCorr (double value) { sumPUPtCorr_  = value; };
         void set_electronPVIndex (int value) { electronPVIndex_  = value; };
         void set_genD0 (double value) { genD0_  = value; };
+        void set_d0SmearingVal (double value) { d0SmearingVal_  = value; };
         void set_passesTightID_noIsolation (const reco::BeamSpot &, const TYPE(primaryvertexs) &, const edm::Handle<vector<reco::Conversion> > &);
         void set_passesVID_vetoID (const bool);
         void set_passesVID_looseID (const bool);
@@ -99,7 +101,8 @@ namespace osu
         double sumChargedHadronPtCorr_;
         int electronPVIndex_;
         double sumPUPtCorr_;
-	double genD0_;
+        double genD0_;
+        double d0SmearingVal_;
         bool passesTightID_noIsolation_;
 
         bool passesVID_vetoID_;
--- a/Collections/interface/Muon.h
+++ b/Collections/interface/Muon.h
@@ -23,14 +23,16 @@ namespace osu
         const double sumChargedHadronPtCorr () const;
         const double sumPUPtCorr () const;
         const int muonPVIndex () const;
-	const double genD0 () const;
+        const double genD0 () const;
+        const double d0SmearingVal () const;
         const bool isTightMuonWRTVtx() const { return isTightMuonWRTVtx_; }
         void set_isTightMuonWRTVtx(const bool isTightMuon);
         void set_pfdBetaIsoCorr (double value) { pfdBetaIsoCorr_  = value; };
         void set_sumChargedHadronPtCorr (double value) { sumChargedHadronPtCorr_  = value; };
         void set_sumPUPtCorr (double value) { sumPUPtCorr_  = value; };
         void set_muonPVIndex (int value) { muonPVIndex_  = value; };
-	void set_genD0 (double value) { genD0_  = value; };
+        void set_genD0 (double value) { genD0_  = value; };
+        void set_d0SmearingVal (double value) { d0SmearingVal_  = value; };
 
 	void set_match_HLT_IsoMu27_v (const bool);
         void set_match_HLT_IsoMu24_v (const bool);
@@ -64,7 +66,8 @@ namespace osu
         double sumChargedHadronPtCorr_;
         int muonPVIndex_;
         double sumPUPtCorr_;
-	double genD0_;
+        double genD0_;
+        double d0SmearingVal_;
 
         double metMinusOnePt_;
         double metMinusOnePx_;
--- a/Collections/plugins/OSUElectronProducer.cc
+++ b/Collections/plugins/OSUElectronProducer.cc
@@ -147,7 +147,10 @@ OSUElectronProducer::produce (edm::Event &event, const edm::EventSetup &setup)
 	}
 
       // produce random d0 value to use in d0 smearing
-      double d0SmearingVal = random->Gaus(0, 0.00142); // todo: remove magic number
+      double d0SmearingVal = 0.0;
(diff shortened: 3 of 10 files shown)


bpf6qc:
Initially I was going to say that we don't store these random-number type things in objects, but I would be wrong: see [the jet producer for smeared JER](https://github.com/OSU-CMS/OSUT3Analysis/blob/master/Collections/plugins/OSUGenericJetProducer.cc#L281). So this is a perfectly fine concept you've laid out here.

For sanity and clarity I would recommend you don't store the "smearingValue" but store the "smearedValue" -- ie (d0 + smearingValue), but that's just me.

Also you are correct about doing it here and not an EventVariableProducer -- an EventVariableProducer stores one value per event in a string, so if you went that route you'd need to have "leadElectronSmearedD0", "sublead...", etc. This is why I say it's correct to do it here.

To answer your questions:
1) You tell from the `event`: `event.isRealData()`
2) Frankly you should have all analyses read a value in. The electron (muon) producer already reads in [this pset](https://github.com/OSU-CMS/OSUT3Analysis/blob/master/Configuration/python/CollectionProducer_cff.py#L56) ([this pset for muons](https://github.com/OSU-CMS/OSUT3Analysis/blob/master/Configuration/python/CollectionProducer_cff.py#L144)) so you could provide some value here. Disappearing tracks already does something like this [here for tracks](https://github.com/OSU-CMS/OSUT3Analysis/blob/master/Configuration/python/CollectionProducer_cff.py#L224), all analyses use this but no one but us cares.
What I recommend is to define a value of 0.0 here for everyone, and then in your protoConfig (or wherever in DisplacedSUSY code) set it to the desired value. This way the value is always set and no one else will need to care. You can see a somewhat convoluted method of this change for DisappTrks [here](https://github.com/OSU-CMS/DisappTrks/blob/f4068d507c13d95b86653423dfa645e6e317ae44/StandardAnalysis/python/utilities.py#L114-L125) -- note the `hasattr` for `OSUTrackProducer` in case that producer isn't being run and the pset wouldn't exist.

jalimena:
thanks, @bpf6qc . i agree with everything. @btcardwell , please follow his suggestions. In particular, I definitely agree about storing the "smearedValue" instead of the "smearingValue".

btcardwell:
Thanks @bpf6qc and @jalimena. Your answers are exactly what I was looking for, and I'll implement them shortly.

Regarding "smearedValue" vs "smearingValue", my initial instinct was the same as both of you, but I wound up storing smearing value for two reasons:
1. We already calculate lepton d0 when making histograms, and it seemed silly to calculate it in yet another place. I know we explicitly decided not to calculate d0 in this way and not in the lepton producer, but I'm having trouble remembering why.
2. Storing the smearing value lets us look at look at the smearing value distribution and confirm it is behaving the way we expect.

I agree storing the smearingValue feels less straightforward on the surface though. Let me know what you guys think.

bpf6qc:
@btcardwell maybe you're talking about the 'WRTPV' calculation? If so then that could be a good reason to keep the 'smearingValue' method. You could always calculate out the 'smearedD0WRTPV' in a similar way and plot the difference to find the correction, but it's really up to you how you implement this. Maybe you could provide a getter function for both the smearingValue and smearedD0.

btcardwell:
I'm referring to [this string](https://github.com/OSU-CMS/OSUT3Analysis/blob/master/Configuration/python/cutUtilities.py#L55), which we use to calculate electron d0 when cutting or plotting.

bpf6qc:
Talking with @jalimena this PR is fine in principle with the work that you've said you would add:
1) Add the same for muons
2) Add a parameter (0.0) for both smearing widths in CollectionProducer_cff.py, then in DisplacedSUSY set it how you please
3) Add an `isRealData()` check to not roll the RNG for real data

Some requests of my own:
4) Name the TRandom3 `rng` as it is in other places (possibly `random` conflicts with some standard libaries)
5) If the smearing width is 0.0, also skip the rolling
6) `delete rng` in the producer's constructor. See OSUGenericJetProducer for how `rng` is used.

btcardwell:
Hi @bpf6qc and @jalimena. I'd like to do a little more testing before merging this PR, but I think everything should be implemented now. I'd be interested in feedback, specifically whether @bpf6qc thinks the rng is being handled properly and if anyone has thoughts on how to avoid the floating point comparison to skip rolling if smearing width is 0.0 ([current naive interpretation](https://github.com/OSU-CMS/OSUT3Analysis/pull/71/commits/38a1f5d0124548f4f9591df91244eaffb2d5a5a2)).

jalimena:
thanks, @btcardwell . looks good from my side. i don't have any good ideas to answer your questions. i think feedback from @bpf6qc  would be good.

bpf6qc:
@btcardwell regarding your issues with the randomness of this parameter (https://cmshead.mps.ohio-state.edu:8080/DisplacedSUSY/1132), I believe I know why this is. If you look at the constructor for TRandom3:

https://root.cern.ch/doc/master/classTRandom3.html#aa0f90fdd325edead6b80afcb05099348

You'll see it comments that a seed argument of 0 will, effectively, give a random seed, however there is a default constructor for seed 4357. 

What I strongly suspect is happening is that when you run over the initial skim, with you say 1000 condor jobs, you create 1000 of the same-seeded RNGs which are then _not_ random between the jobs -- however in the larger statistics of the initial skim, you see the distribution play out okay (it's not actually okay if you look. In the skim-of-skim however you have fewer events, so each job won't populate the distribution, and what you're seeing is the same exact partially-filled distribution added to itself N-many times.

The solution to try: just call `rng = new TRandom3(0);`. In fact if this fixes your ELOG issue, if you see any other `TRandom3()` instances then this is a bug we have never noticed before, and should be fixed.

I'll give other comments on your commits soon.

bpf6qc:
Some comments on the current status:

1) See previous comment about TRandom3(0). If this fixes your ELOG issue, apply this to the JER smearing because we would know it's wrong.
2) How about instead of a width of 0.0 meaning we turn this off, use a negative value (-1.0). This avoids your float comparison issue.
3) If you're skipping the smearing, maybe we should skip creating the RNG completely. As in something like...
```
if(d0SmearingWidth_ >= 0) rng = new TRandom3(0);
```
...
```
OSUElectronProducer::~OSUElectronProducer ()
{
  if(rng) delete rng;
}
```
This last point is super marginal as the TRandom3 object is small.

btcardwell:
Alright, I believe this PR is ready to be merged. @bpf6qc, you were spot on with the TRandom3 constructor being off.

btcardwell:
And I don't see any other instances of TRandom3() in OSUT3Analysis. Specifically, JER already uses the [correct constructor](https://github.com/OSU-CMS/OSUT3Analysis/blob/e4e83847256c5b8dc1b13796ab58f40d644c07f9/Collections/plugins/OSUGenericJetProducer.cc#L87).

PR #72, opened 2018-08-24
Tweaks for LPC

bpf6qc:
Limits multithreading to only 4 merge processes so that LPC admins don’t yell at you
Adds a flag to skip the actual merging of hist_*.root files, so that you can quickly check skim files and create resubmission scripts


bpf6qc:
@jalimena I've been running this version for a while and the flag does work, and I can see in `top` that it only spawns 4 merging threads; I haven't been yelled at by admins yet.

PR #75, opened 2018-09-13
add HardInteractionMcparticles

jalimena:
adds HardInteractionMcparticles, which is needed to be able to use the framework fully with the hardInteractionMcparticles collection.


jalimena:
this PR now finally does everything i need it to do, and I've tested that it works properly on 10 generated events. I'm going to test it now on a few more events, and then it should be good to go.

PR #76, opened 2018-09-20
Added DYJetsToLL_5to50.

aehart:


jalimena:
Displaced leptons is still using a DYJetsToLL_10to50 sample for 80X and 94X, so removing this from the composite dataset definition on L418 may not be appropriate.

Other than that - do you need to add the cross section?

aehart:
OK, I'll change the composite dataset definition back. We can just change that in our local config files.

And yeah, I guess this is the place to put the cross section. In the past, we always added them to the database, but I think all our datasets will be not registered in the database now, so I'll add it here.

btcardwell:
Yup, looks good to me

PR #78, opened 2018-09-25
Update 2017 fiducial maps for ntuples

bpf6qc:
See:
https://cmshead.mps.ohio-state.edu:8080/DisappearingTracks/985
https://cmshead.mps.ohio-state.edu:8080/DisappearingTracks/987

Muons are actually only B-E, F will be added when ntuples are complete.


bpf6qc:
Actually, hold on this PR for a moment. I'm just checking the method for something we might have neglected.

bpf6qc:
Talking it through with Andrew, the way we calculate this is still correct. So these files are what we want to commit.

PR #79, opened 2018-10-04
Changed how x509 proxies are handled when at Rutgers.

aehart:
The usual options for x509 proxies do not work at the Rutgers T3 because HTCondor does not have permission to read the proxy file, for some reason.

So now osusub.py copies the proxy file to the condor directory, relaxes the permissions, and transfers it to the worker node like a regular file. The permissions are then restricted after submitting the jobs.

Since this is a mild security hole, this is only done at Rutgers. The behavior should be unchanged when running anywhere else.

The Rutgers xrootd redirector was also added as an argument to the redirector option.

### DIFF
--- a/DBTools/scripts/osusub.py
+++ b/DBTools/scripts/osusub.py
@@ -412,7 +412,7 @@ def MakeCondorSubmitScript(Dataset,NumberOfJobs,Directory,Label, SkimChannelName
             SubmitFile.write('should_transfer_files   = YES\n')
             SubmitFile.write('Transfer_Input_files = ' + FilesToTransfer + '\n')
             if UseGridProxy and not rutgers:
-                SubmitFile.write('x509userproxy = ' + userProxy + '\n')
+                SubmitFile.write('x509userproxy = /tmp/' + proxy + '\n')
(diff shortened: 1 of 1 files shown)


aehart:
Thanks for taking a look @jalimena .

The "Rutgers" on L123 is a key to the RedirectorDic. It determines the appropriate argument to the redirector option (i.e., ... --redirector=Rutgers) and is not used explicitly in the rest of the code.

The "rutgers" used elsewhere is the name of a variable defined on L1192.

I have been using this branch at the Rutgers T3 for some time, and have been able to submit jobs that successfully run over remote datasets with xrootd. I was careful to quarantine the changes in how x509 proxies are handled so they only take effect when run from Rutgers, but I can test that the script still runs as expected at OSU today.

jalimena:
ohhh, i missed L1192, sorry about that. yeah, just a quick check on OSU would convince me we can merge.

aehart:
There was one line where I forgot to update a variable name. I fixed that, and now it runs fine at OSU (see output for job 0 in /data/users/hart/condor/rutgers_test/MET_2017B/).

PR #82, opened 2018-10-30
Handle secondary input files for remote running over empty skims

bpf6qc:
Currently SkimModifier calls getOriginalFile(s) to figure out the list of secondary input files, in cases where you're submitting a job over an empty skim. This function tries to open a local file `SkimDirectory.txt` to know where the original non-empty skim is stored. If you're submitting a job over a remote skim, ie `-s root://...`, then we never transferred `SkimDirectory.txt`.

You could `xrdcp` this file and figure things out, but then `SkimDirectory.txt` is a path relative within the remote machine; it could be `/eos/uscms/store/...` on LPC or `/cms/ahart/disappearingTracks/...` on Rutgers...and you in general don't know the physical path on a machine from the xrootd redirector. You'd need to know that the LPC strips `/eos/uscms`, Rutgers changed `/cms` to `/store/user`, and etc for special cases.

...Or you could be lazy and use the fact that the datasetInfo already has a complete history of the skim inputs. If the input skim's datasetInfo has secondary input files, you need to keep that the same, and if there were none then you just want the listOfFiles for the input skim.

This works when running over things stored at the LPC because that natively uses xrootd, but this won't work for OSU or Rutgers' T3 because those paths will still be local physical paths.

### DIFF
--- a/DBTools/scripts/osusub.py
+++ b/DBTools/scripts/osusub.py
@@ -1015,20 +1015,23 @@ def SkimModifier(Label, Directory, crossSection, isRemote = False):
             add += '"file:' + os.path.realpath (s) + '",\n'
     add += ']\n'
 
-    add += 'listOfSecondaryFiles = [\n'
-    tempdir = tempfile.mkdtemp ()
-    sys.path.append (tempdir)
-    cwd = os.getcwd ()
-    os.chdir (tempdir)
-    if lpcCAF:
-        for s in getOriginalFiles (skimFiles, tempdir):
-            add += '"' + re.sub (r"/eos/uscms/store/", r"root://cmseos.fnal.gov//store/", os.path.realpath (s)) + '",\n'
+    if isRemote:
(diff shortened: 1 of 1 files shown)


jalimena:
hi @bpf6qc , does this PR happen to fix the problem Bryan saw yesterday or the day before?

in any case, this PR seems good. Have you tested it yet?

bpf6qc:
@jalimena in the case of running at the OSU T3, using skim files on the LPC, this will fix it. Take a look at:
`/data/users/bfrancis/condor/testSubmission2/SingleEle_2017B/datasetInfo_SingleEle_2017B_cfg.py`
for a successful run (I ran only 1 of these jobs but it finished successfully).

What it won't fix however is if the skim is stored anywhere but the LPC, so it's up to reviewers if that's acceptable for now.

jalimena:
thanks, Brian. I think this is acceptable for now, we can always make more developments later.

Maybe @aehart  and @btcardwell  can comment if they think this PR is ready to be merged? From my side, it is.

btcardwell:
Yes, I think we can merge this PR as is and develop something for other remote skims if needed in the future.

PR #83, opened 2018-11-06
Fix typo in replaceSingleCut

bpf6qc:
Apparently no one uses this function, simple typo/bug fix.


jalimena:
seems like a no-brainer, and also this is just a print statement. done.

PR #85, opened 2018-11-16
Use xrootd paths as-is for secondary file list

bpf6qc:
Running over empty skims on the LPC was broken because `os.path.realpath` was being called on xrootd paths, which yields bad paths which fail the job. Example:

`root://cmseos.fnal.gov//store/user/lpclonglived/DisappTrks/skim/2017/electronTagSkim/SingleEle_2017B/ElectronTagSkim/skim_462.root`
gives something like
`/tmp/tmpQUlSst/root:/cmseos.fnal.gov/store/user/lpclonglived/DisappTrks/skim/2017/electronTagSkim/SingleEle_2017B/ElectronTagSkim/skim_462.root`
because `:` is interpreted as a separator for multiple files, and `//` is stripped to `/`.

So the solution is to take the file path itself if it starts with `root://` instead of calling `realpath`.

Testing: datasetInfo_*_cfg.py are no longer filled with garbage paths and jobs are running successfully @ LPC.

### DIFF
--- a/DBTools/scripts/osusub.py
+++ b/DBTools/scripts/osusub.py
@@ -1025,7 +1025,10 @@ def SkimModifier(Label, Directory, crossSection, isRemote = False):
         os.chdir (tempdir)
         if lpcCAF:
             for s in getOriginalFiles (skimFiles, tempdir):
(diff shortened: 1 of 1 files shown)


aehart:
Good catch and your solution works, so I'll go ahead and merge it.

PR #88, opened 2019-01-03
Added message to setupFramework.py about necessary merge-topic for electrons.

aehart:
Unless I'm missing something, the V2 file is only available in CMSSW_10_2_0 and later, so we should use the V1 file in CMSSW_9_4_X.

### DIFF
--- a/AnaTools/scripts/setupFramework.py
+++ b/AnaTools/scripts/setupFramework.py
@@ -4,6 +4,29 @@ import sys
 import os
 from optparse import OptionParser
 
+# define some constants with meaningful names for the ANSI color codes
+# https://en.wikipedia.org/wiki/ANSI_escape_code#Colors
+
+A_BLACK           =  "\033[30m"
+A_RED             =  "\033[31m"
+A_GREEN           =  "\033[32m"
+A_YELLOW          =  "\033[33m"
+A_BLUE            =  "\033[34m"
+A_MAGENTA         =  "\033[35m"
+A_CYAN            =  "\033[36m"
+A_WHITE           =  "\033[37m"
+
+A_BRIGHT_BLACK    =  "\033[1;30m"
+A_BRIGHT_RED      =  "\033[1;31m"
+A_BRIGHT_GREEN    =  "\033[1;32m"
+A_BRIGHT_YELLOW   =  "\033[1;33m"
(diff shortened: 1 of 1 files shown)


jalimena:
Hi @aehart , when I last looked at the egamma recipe twiki, there was a recipe that they advocated that you use in 94X when X>=9 and you want the V2 VID. Now when I look, I can find this recipe for 2017 VIDs:
https://twiki.cern.ch/twiki/bin/view/CMS/EgammaMiniAODV2#2017_MiniAOD_V2
Which I think also corresponds to V2. So, it's a little confusing, but I'm not sure you're right.

bpf6qc:
In https://github.com/OSU-CMS/OSUT3Analysis/pull/86 @jalimena commented what the recipe was to add the V2 configuration file, but we left V1 as the default in 94X releases until analyses were ready to change. I'm kind of confused about this V2 business now though -- I don't see in this twiki link the function calls I would expect from the previous recipe:
`switchOnVIDElectronIdProducer(process, DataFormat.MiniAOD)` etc...
So my confusion is if the recipe has changed entirely, or this link @jalimena sends is only referencing a bug-fix.

Also I see an intriguing thing here:
https://twiki.cern.ch/twiki/bin/view/CMS/EgammaMiniAODV2#Accessing_ID_result
Which if they've added these methods to standard pat::Electron, that would actually solve some severe headaches we've had in this.

I'd like to look into this more before we change things. For now I've just added to the Disappearing Tracks recipe:
`git cms-merge-topic UAEDF-tomc:eleCutBasedId_94X_V2`
which solves the import error @aehart is probably talking about and still keeps V1 as default.

aehart:
Yeah, this PR was my clever way of asking what merge-topics I needed. Thanks @jalimena and @bpf6qc .

Just to make the workflow a little more fool-proof, I added a warning to setupFramework.py. That's now the content of this PR. It may be good practice to add similar warnings here for other merge-topics that may be required for various things.

aehart:
Good idea. Name of the PR has been updated.

jalimena:
looks good to me. @bpf6qc ? can we merge?

bpf6qc:
@jalimena this looks good to me; I'll make an issue to remind us to look at the VID recipe too.

PR #89, opened 2019-01-07
Added a histogram of the number of layers with measurements for tracks.

aehart:
Just a small (20 bins) histogram that should be in the standard set of track histograms.


bpf6qc:
Looks good for me, good change.

PR #92, opened 2019-01-24
Fix bug causing full DAS dataset to be used as secondary files when r…

bpf6qc:
…unning remotely over non-empty skims


jalimena:
+1

btcardwell:
Looks straightforward and good

aehart:
I requested separately that @bpf6qc fix the obvious syntax error caused by the single quotes and switch to using `os.path.basename`, since that's a little more transparent.

Now, it works and looks good, so I'll merge.

PR #94, opened 2019-01-25
LPC condor refactor

aehart:
A few changes to deal with the LPC's condor refactor:
- Always use a grid proxy when submitting jobs at the LPC
- Allow the grid proxy to be in locations other than /tmp/, as determined by X509_USER_PROXY
- Remove `Getenv = True` from `condor.sub` when submitting jobs at the LPC
- Added a warning about submitting from SL6 at the LPC

Jobs seem to run just fine when submitted from an SL7 release on an SL7 node. But right now, success varies wildly when submitting from SL6. This is likely an issue that the LPC admins will have to address.

But for us, it's easy enough to switch to SL7, so I just added a warning when submitting jobs from SL6, informing the user that they should switch.

### DIFF
--- a/Configuration/python/Color.py
+++ b/Configuration/python/Color.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python
+
+# define some constants with meaningful names for the ANSI color codes
+# https://en.wikipedia.org/wiki/ANSI_escape_code#Colors
+
+A_BLACK           =  "\033[30m"
+A_RED             =  "\033[31m"
+A_GREEN           =  "\033[32m"
+A_YELLOW          =  "\033[33m"
+A_BLUE            =  "\033[34m"
+A_MAGENTA         =  "\033[35m"
+A_CYAN            =  "\033[36m"
+A_WHITE           =  "\033[37m"
+
+A_BRIGHT_BLACK    =  "\033[1;30m"
+A_BRIGHT_RED      =  "\033[1;31m"
+A_BRIGHT_GREEN    =  "\033[1;32m"
+A_BRIGHT_YELLOW   =  "\033[1;33m"
+A_BRIGHT_BLUE     =  "\033[1;34m"
+A_BRIGHT_MAGENTA  =  "\033[1;35m"
+A_BRIGHT_CYAN     =  "\033[1;36m"
+A_BRIGHT_WHITE    =  "\033[1;37m"
+
+A_RESET           =  "\033[0m"
--- a/DBTools/scripts/osusub.py
+++ b/DBTools/scripts/osusub.py
@@ -402,17 +402,17 @@ def MakeCondorSubmitScript(Dataset,NumberOfJobs,Directory,Label, SkimChannelName
                 FilesToTransfer += ',datasetInfo_' + Label + '_cfg.py'
             if UseGridProxy:
                 if rutgers:
-                    shutil.copy ("/tmp/" + proxy, Directory + "/" + proxy)
-                    os.chmod (Directory + "/" + proxy, 0644)
-                    FilesToTransfer += ',' + proxy
+                    shutil.copy (proxy, Directory + "/" + os.path.basename (proxy))
+                    os.chmod (Directory + "/" + os.path.basename (proxy), 0644)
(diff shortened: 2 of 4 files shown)


Configuration/python/Color.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
```
bpf6qc: There are more instances of these that could be removed now with this module:
```
Configuration/python/ProgressIndicator.py
AnaTools/interface/DataFormat.h
AnaTools/scripts/setupFramework.py
```
Good idea, please clean those 3 up too.

DBTools/scripts/osusub.py
```diff
@@ -1189,27 +1203,22 @@ def copyDatasetInfoToTempFile (f, tempdir):
...
     print
-    print "Warning! You are working on the LPC, but have not provided a \"Hadoop\" directory for your skim with \"-H\"."
-    print "         LPC's condor nodes have no access to regular directories, so you will never be able to run over"
-    print "         the resulting skim files using condor, nor will you be able to access them with xrootd remotely."
-    print "         Consider using -H with eos!"
+    print A_BRIGHT_RED + "Warning! You are working on the LPC, but have not provided a \"Hadoop\" directory"
+    print                "         for your skim with \"-H\". Your skim files will be written directly to"
+    print                "         your condor directory, which may threaten the quota on your home"
+    print                "         directory. Consider using -H with eos!" + A_RESET
     print
```
bpf6qc: +1 for remembering this

DBTools/scripts/osusub.py
```diff
@@ -1338,12 +1347,12 @@ def copyDatasetInfoToTempFile (f, tempdir):
                     os.syetem('./lxbatchSub.sh')
                 else:
                     os.symlink ("../" + os.environ["CMSSW_VERSION"] + ".tar.gz", os.environ["CMSSW_VERSION"] + ".tar.gz")
-                    cmd = "condor_submit condor.sub"
-                    if os.path.isfile (proxy):
-                        os.chmod (proxy, 0644)
+                    cmd = "LD_LIBRARY_PATH= condor_submit condor.sub"
```
bpf6qc: I don't understand the `LD_LIBRARY_PATH= ` part. Why would you export an environment variable instead of executing the submit command?

Configuration/python/Color.py
aehart: I imported `Color.py` in `ProgressIndicator.py`.

But `DataFormat.h` is C++, not Python, so I can't import there. And `setupFramework.py` is typically run before doing `scram b`, which sets up the links in `$CMSSW_BASE/python/`, so it cannot import things in the framework, unless we want the user to run `scram b` first, then `setupFramework.py`, then run `scram b` again. Which I don't think we do.

DBTools/scripts/osusub.py
aehart: This may be some quirk of my environment, though I have no idea where it's coming from. When I try to run any of the condor commands after doing `cmsenv`, I get the following error:
```
Traceback (most recent call last):
  File "/usr/local/bin/condor_q", line 6, in <module>
    import argparse
  File "/usr/lib/python2.7/argparse.py", line 85, in <module>
    import collections as _collections
  File "/usr/lib/python2.7/collections.py", line 8, in <module>
    from _collections import deque, defaultdict
ImportError: /usr/lib/python2.7/lib-dynload/_collectionsmodule.so: wrong ELF class: ELFCLASS32
```
It only goes away if I unset `LD_LIBRARY_PATH`.

This line isn't exporting `LD_LIBRARY_PATH`. It is running `condor_submit condor.sub` with `LD_LIBRARY_PATH` set to nothing. This is different than exporting because the change to the environment persists only for the execution of the command.

Like I said, this is probably just a quirk of my environment, but unsetting `LD_LIBRARY_PATH` when executing condor commands in `osusub.py` shouldn't hurt anything for anyone else.

Configuration/python/Color.py
bpf6qc: Whoops, fair points.

PR #95, opened 2019-01-25
Fix checkJobs for different log names

bpf6qc:
A future commit will create condor log files like
`condor_SingleEle_2017B_7.log`
and checkJobs currently expects only e.g. `condor_7.log`. This prevents a script crash and checks these new logs.


aehart:
Looks good and works. Merging now.

PR #96, opened 2019-01-27
Fix merging lpc

bpf6qc:
These changes completely ignore ongoing work with https://github.com/OSU-CMS/OSUT3Analysis/pull/91, but merging on LPC when the skim is on EOS is currently broken and these changes restore the ability to merge there. They've now disallowed all standard disk operations on EOS (ls, glob.glob, listdir, etc) so only xrootd is available.

The other PR will be built on top of these necessary changes soon.

### DIFF
--- a/Configuration/python/mergeUtilities.py
+++ b/Configuration/python/mergeUtilities.py
@@ -10,6 +10,7 @@ import pickle
 import shutil
 import math
 import socket
+import tempfile
 from threading import Thread, Lock, Semaphore
 from multiprocessing import cpu_count
 from OSUT3Analysis.Configuration.configurationOptions import *
@@ -17,7 +18,7 @@ from OSUT3Analysis.Configuration.processingUtilities import *
 from OSUT3Analysis.Configuration.formattingUtilities import *
 from OSUT3Analysis.DBTools.condorSubArgumentsSet import *
 import FWCore.ParameterSet.Config as cms
-from ROOT import TFile
+from ROOT import TFile, TNetXNGFile
 
 NTHREADS_FOR_BATCH_MERGING = 1
 
@@ -86,6 +87,7 @@ def MessageDecoder(Message, Good):
 ###############################################################################
 def GetGoodRootFiles(Index):
     return " ".join (glob.glob ("*_" + str (Index) + ".root"))
+
 def MakeWeightsString(Weight,FilesSet):
     Str = ''
     for i in range(0,len(FilesSet)):
@@ -94,6 +96,7 @@ def MakeWeightsString(Weight,FilesSet):
         else:
             Str = Str + ',' + str(Weight)
     return Str
+
 ###############################################################################
 #   Get the total number of events from cutFlows to calculate the weights     #
 ###############################################################################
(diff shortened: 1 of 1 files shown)


jalimena:
I've tested this in two ways, and in one I'm getting an error.

First, I ran some quick jobs on the LPC, where the input files were miniAOD. Then I successfully merged them. The output can be found here:
`/eos/uscms/store/group/lpclonglived/DisplacedLeptons/testPR96_EMuPreselection_2016/
/uscms_data/d3/alimena/condor/testPR96_EMuPreselection_2016`

Then I tried to run a skim over these output files, but I get an error on submission. Here is the command I'm using and the log output I get:
`susub.py -l PreselectionOptions.py -w testPR96_EMuPromptControlRegion_2016  -H /eos/uscms/store/group/lpclonglived/DisplacedLeptons -a Preselection -t OSUT3Ntuple -s testPR96_EMuPreselection_2016
using 80X samples
normalizing MC to 16146.2 1/pb

Warning! You are working in SL6 on the LPC. Job submission is currently
         problematic in SLF6, and you are encouraged to switch to SL7.

# Conditions read from  CMS_CONDITIONS  via FrontierProd 
using global tag 80X_mcRun2_asymptotic_2016_TrancheIV_v8...
# EventWeights applied: 2016
# Lepton SFs: 2016
# Using 2016 triggers
You have specified a skim as input.  Will obtain cross sections for dataset stop900_10mm from the database.
Traceback (most recent call last):
  File "/uscms_data/d3/alimena/DisplacedLeptons/CMSSW_8_0_31/bin/slc6_amd64_gcc530/osusub.py", line 1303, in <module>
    DatasetRead = MakeFileList(DatasetName,arguments.FileType,WorkDir,dataset, UseAAA, crossSection)
  File "/uscms_data/d3/alimena/DisplacedLeptons/CMSSW_8_0_31/bin/slc6_amd64_gcc530/osusub.py", line 922, in MakeFileList
    exec('import datasetInfo_' + Label +'_cfg as datasetInfo')
  File "<string>", line 1, in <module>
  File "/uscms_data/d3/alimena/DisplacedLeptons/CMSSW_8_0_31/src/DisplacedSUSY/EMuChannel/test/condor/testPR96_EMuPromptControlRegion_2016/stop900_10mm/datasetInfo_stop900_10mm_cfg.py", line 18
    
                                         ^
SyntaxError: invalid syntax
`

Is the problem me or this PR?

bpf6qc:
The immediate problem is a lack of documentation, but there's also one remaining problem we need to fix. Both are actually in osusub.py and not the merging in this PR. EOS is no longer mounted **at all** except for xrootd. You can see in
```/uscms_data/d3/alimena/condor/testPR96_EMuPreselection_2016/stop900_10mm/datasetInfo_stop900_10mm_cfg.py```
that your second submission stopped when trying to get the number of skim files being used as the input. This is because you used `-s /eos/uscms/<etc>` as a real path, and not an xrootd path. You should also see something like `permission denied` at some point in your output, because this is the error.

I can see that your merging did work:
```
xrdfsls /eos/uscms/store/group/lpclonglived/DisplacedLeptons/testPR96_EMuPreselection_2016/stop900_10mm/Preselection/ # these text files are the result of merging
xrdfsls /eos/uscms/store/group/lpclonglived/DisplacedLeptons/testPR96_EMuPreselection_2016/stop900_10mm/Preselection/SkimNumberOfEvents.txt
```

The remaining problem is that osusub.py can't understand /eos/uscms paths in the `-H` argument. It knows how to turn that into an xrootd path, but it still tries to look at the path on disk.

TL;DR
1) LPC EOS is **always** remote now, so use `-s root://cmseos.fnal.gov//store/group/lpclonglived/DisplacedLeptons/testPR96_EMuPreselection_2016`
2) `-H` is still broken in a similar way but I know how to fix this.

jalimena:
Ok so what you're saying we need to do is:
1. Improve the documentation, say with an OSUT3Analysis elog post.
2. Fix the problem in `-H` --> should this happen in this PR or in another one?

Do you have more changes that you want to push to this PR, or should we merge this one already?

bpf6qc:
This PR won't get any more commits, we've been able to merge with it so it fixes the issue at hand. The other active PR is actually improvements to merging, and I'll continue to work on that. My thinking is this should be merged to fix the issue, and I'll get another one going for the separate issue in osusub.py.

jalimena:
makes sense. done.

PR #100, opened 2019-02-07
Update jer methods

bpf6qc:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/JetResolution#Smearing_procedures

The prescription from JME has changed, and these are implemented here. There is a flag available to toggle the old behavior, but the default is to use the previous method in 2015 and new method in 2016-8.

Also the flag to use JER from the global tag was only set to true for 2016, so 2017 users were using 2015 text file payloads for JER. The GT is now used in 2016-8.

### DIFF
--- a/AnaTools/interface/CommonUtils.h
+++ b/AnaTools/interface/CommonUtils.h
@@ -272,11 +272,15 @@ anatools::getObjectHash(const T& object){
 template<class T> bool
 anatools::jetPassesTightLepVeto (const T &jet)
 {
-  // Taken from recommendations from the JetMET group for 13 TeV:
+#if CMSSW_VERSION_CODE >= CMSSW_VERSION(9,4,0)
+  // https://twiki.cern.ch/twiki/bin/view/CMS/JetID13TeVRun2017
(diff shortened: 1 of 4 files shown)


aehart:
There seem to be a couple disagreements with what is written on the TWiki:
- In the case that no matching gen-jet is found, the TWiki doesn't require the scale factor to be greater than one. The momentum would just not be rescaled, not set to an invalid value. Granted, the scale factors on the TWiki are almost all greater than one, so it doesn't really matter.
- The sigma for the Gaussian on the TWiki is just the jet energy resolution. Then the random number is multiplied by the expression under the radical.

bpf6qc:
@aehart updated for your comment. The JME twiki's language has always said that a scale factor of less than one made the stochastic/random method impossible, and I believe they used to say that such jets were nonsense and shouldn't be used. That language is gone now so it now just doesn't scale it.

Secondly the example code the twiki links to is also wrong apparently, as I just copy-pasted from there. But you're right, the twiki says otherwise. This PR now does what the twiki itself says.

jalimena:
Looks fine from my perspective. Displaced leptons doesn't use jets, so I can't foresee a useful test from my side. I suggest that Brian and/or Andrew test it, and when you are satisfied, merge.

aehart:
@bpf6qc thanks for the changes. Looks good now.

PR #102, opened 2019-02-12
Weight histograms using "optional_dict_ntupleEff".

aehart:
The user can include an optional dictionary called "optional_dict_ntupleEff" in their local config file, with dataset names as keys and floating point numbers as values. The values are used to weight the corresponding datasets during merging.

The intended use case of this dictionary is to provide skim efficiencies for skims that are produced outside the usual skim framework, e.g., during ntuple making.

There is no change in merging behavior if no "optional_dict_ntupleEff" is provided.

### DIFF
--- a/Configuration/python/mergeUtilities.py
+++ b/Configuration/python/mergeUtilities.py
@@ -56,7 +56,7 @@ def MakeSubmissionScriptForMerging(Directory, currentCondorSubArgumentsSet, spli
 ###############################################################################
 #                 Make the configuration for condor to run over.              #
 ###############################################################################
-def MakeMergingConfigForCondor(Directory, OutputDirectory, split_datasets, IntLumi):
+def MakeMergingConfigForCondor(Directory, OutputDirectory, split_datasets, IntLumi, optional_dict_ntupleEff):
     MergeScript = open(Directory + '/merge.py','w')
     MergeScript.write('#!/usr/bin/env python\n')
     MergeScript.write('from OSUT3Analysis.Configuration.mergeUtilities import *\n')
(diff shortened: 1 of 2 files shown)


bpf6qc:
Addresses the problem, looks good.

jalimena:
Can @aehart make an elog to describe when you would want to use this? I'm not sure I understand if this would affect displaced leptons or not.

bpf6qc:
@jalimena the ELOG is still unavailable (I need to go yell at IT again...) so I don't think he can, really. The problem we ran into with disappearing tracks is that when we produce nTuples, our extended MINIAOD format, we also apply a skim. This isn't done in the framework, so OSUT3Analysis never knew about it. It just looks at the size of the dataset you're using on DAS, and then says that's the nGen you want to use in cross section scaling.

A blown-up example: say you skim a QCD sample with CRAB and only select 1 event. When you use the framework to run over that published dataset, it thinks the entire MC is just 1 event, so it scales it by `lumi * xsec / 1`. But you've missed the millions of events in the full dataset. You need to provide information about the efficiency of the skim step.

bpf6qc:
You've already done that effectively with the EM enriched QCD samples; there is a filter efficiency that you've provided in the crossSection field, `xsec * filterEff`. Ours is not really a filter efficiency in the same way so this is another method.

So to answer: you would only create this `optional_dict_ntupleEff` dictionary if you did a private skim in CRAB, without the framework. I don't believe displaced leptons do this, so your group would never need this.

jalimena:
ok thanks for explaining! indeed, we have avoided using CRAB for displaced leptons, so we're unaffected by this PR

PR #103, opened 2019-02-14
add 2018 jsons to osusub.py

jalimena:
Adds 2018 JSON files to osusub.py. Right now, there is only the prompt reco file available, so we should use "P_Golden18" for example.


jalimena:
tested in CMSSW_9_4_8 with this command:
`osusub.py -P -A -l SkimOptions.py -w EMuInitialSkim_2018Analysis_14Feb2019 -H /eos/uscms/store/group/lpclonglived/DisplacedLeptons -J P_Golden18`

Successfully obtained the 2018 json file, see e.g.
`/uscms_data/d3/alimena/condor/EMuInitialSkim_2018Analysis_14Feb2019/MuonEG_2017C/Cert_314472-325175_13TeV_PromptReco_Collisions18_JSON.txt`

Notice that I ran over 2017 data because that was all set up, and we are in the process of setting up 2018 data fully.

PR #105, opened 2019-02-18
Add flag to ISRWeightProducer to account for Pythia8 behavior

bpf6qc:
Also remove the recursion in isOriginalParticle from LifetimeWeightProducer to be the same as ISRWeightProducer.

### DIFF
--- a/AnaTools/plugins/ISRWeightProducer.cc
+++ b/AnaTools/plugins/ISRWeightProducer.cc
@@ -4,11 +4,12 @@
 
 ISRWeightProducer::ISRWeightProducer (const edm::ParameterSet &cfg) :
   EventVariableProducer(cfg),
-  pdgIds_            (cfg.getParameter<vector<int> > ("pdgIds")),
-  motherIdsToReject_ (cfg.getParameter<vector<int> > ("motherIdsToReject")),
-  weightFile_        (cfg.getParameter<string> ("weightFile")),
-  weightHist_        (cfg.getParameter<vector<string> > ("weightHist")),
-  weights_           ({})
+  pdgIds_                  (cfg.getParameter<vector<int> >    ("pdgIds")),
+  motherIdsToReject_       (cfg.getParameter<vector<int> >    ("motherIdsToReject")),
+  requireLastNotFirstCopy_ (cfg.getParameter<bool>            ("requireLastNotFirstCopy")),
+  weightFile_              (cfg.getParameter<string>          ("weightFile")),
+  weightHist_              (cfg.getParameter<vector<string> > ("weightHist")),
+  weights_                 ({})
 {
   mcparticlesToken_ = consumes<vector<TYPE(hardInteractionMcparticles)> > (collections_.getParameter<edm::InputTag> ("hardInteractionMcparticles"));
 }
@@ -68,6 +69,16 @@ ISRWeightProducer::AddVariables (const edm::Event &event) {
   double py = 0.0;
 
   for(const auto &mcparticle : *mcparticles) {
+
+    // Pythia8 first creates particles in the frame of the interaction, and only boosts them 
+    // for ISR recoil in later steps. So only the last copy is desired. A particle that's both 
+    // a last and first copy means it is a decay product created after the boost is applied,
+    // e.g. a neutralino from a chargino decay, and we've already added the mother chargino.
+    if(requireLastNotFirstCopy_) {
+      if(!mcparticle.isLastCopy()) continue;
+      if(mcparticle.statusFlags().isFirstCopy()) continue;
+    }
+
     for(const auto &pdgId : pdgIds_) {
       if(mcparticle.numberOfMothers() && 
          !mcparticle.motherRef().isNull() && 
@@ -75,11 +86,13 @@ ISRWeightProducer::AddVariables (const edm::Event &event) {
         continue;
       }
 
-      if(abs(mcparticle.pdgId()) == abs(pdgId) && isOriginalParticle(mcparticle, mcparticle.pdgId())) {
+      if(abs(mcparticle.pdgId()) == abs(pdgId) && 
+         (requireLastNotFirstCopy_ || isOriginalParticle(mcparticle, mcparticle.pdgId()))) {
         px += mcparticle.px();
         py += mcparticle.py();
       }
     }
+
   }
 
   double pt = sqrt(px*px + py*py);
@@ -107,6 +120,7 @@ ISRWeightProducer::AddVariables (const edm::Event &event) {
 #endif
 }
 
+// isOriginalParticle: particle is "original" if its mother has a different pdgId
 bool
 ISRWeightProducer::isOriginalParticle (const TYPE(hardInteractionMcparticles) &mcparticle, const int pdgId) const
 {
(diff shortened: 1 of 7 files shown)


aehart:
In the LifetimeWeightProducer, I think we do _not_ want the "original" particle. We want the last copy, just like in the ISRWeightProducer. Otherwise, when we calculate c*tau, we aren't using the correct boost for the version of the chargino that actually propagates.

bpf6qc:
I did check LifetimeWeightProducer despite the few changes in this PR. It was not obvious what was happening, because I did not know you could iterate over daughters with
`for (const auto &daughter : mcparticle)`

But with that, `getFinalPosition` actually walks down the decay until it reaches a particle without any daughters, so you actually do get the correct cTau starting from the "original" particle -- starting from the lastCopy chargino would not change the result, as long as you can never use multiple copies of the same chargino. 

I considered changing the logic of LifetimeWeightProducer to instead get the lastCopy particle and then find the vertex of one of its daughter, but that seemed risky seeing as it works.

bpf6qc:
@aehart the one cute idea I had on my home however was that we could have LifetimeWeightProducer produce multiple weights for a range of possible destination ctaus. So you run over 100cm and this produces
```100cm * range(2, 10)```

and if you then make a histogram of all of these (binned properly finely), you then could reweight 100cm to all of those from the same file. You might never need to run over the reweighted samples again.

**Comment by aehart** [comment]
The problem is that it calculates the boost factor using the four-vector of the original particle. As we've seen, the four-vector of the original particle is different than that of the propagating particle in our Pythia8 samples, and we need to use the latter to calculate the boost.

So I think we need `getFinalPosition` to output not only the vertex of the first decay product, but also the four-vector of the decay product's mother. Then use that four-vector in `getCTau` to calculate the boost.

Your other idea is cute, and would help reduce the number of jobs we have to run to get signal acceptances by an order of magnitude, but for the moment, I think we should just get the calculation of c*tau correct.

**Comment by bpf6qc** [comment]
The last commit provides this `lastNotFirstCopy` flag to LifetimeWeightProducer as well, and this is observed to fix the issue and the LastCopy chargino is used for the boost.

Also a small change to smeared jet pt -- by storing the regular jet pt when there are no genJets available, this makes it possible to just always use the smearedPt as your cut and it will work in both data and MC.

Lastly and most significantly this vectorizes LifetimeWeightProducer, and removes the reweighted AMSB datasets from existence. Associated changes will need to be made in the DisappTrks repo to accommodate this, but basically if you run a sample with ctau = 100cm, the following weights are produced (debug output):
```
injected lifetimeWeight_1000024_1000cmTo200cm = 3.34709
injected lifetimeWeight_1000024_1000cmTo300cm = 2.63756
injected lifetimeWeight_1000024_1000cmTo400cm = 2.15068
injected lifetimeWeight_1000024_1000cmTo500cm = 1.80907
injected lifetimeWeight_1000024_1000cmTo600cm = 1.55883
injected lifetimeWeight_1000024_1000cmTo700cm = 1.36844
injected lifetimeWeight_1000024_1000cmTo800cm = 1.21903
injected lifetimeWeight_1000024_1000cmTo900cm = 1.09879
```
What this allows is to run the sample (100cm, unreweighted) without _any_ lifetime weights applied. Then as long as you histogram these eventvariables with sufficiently fine-binning, such that the binning error on the integral of said histogram is small, then you can adjust the signal yield for the integral of the weights. This saves you from running of 9 reweighted samples.

**Comment by bpf6qc** [comment]
Some debugging to test, running over one event in AMSB 800GeV_200cm skim acceptance results with one chargino with ctau = 100.337cm. I manually set the rules in LifetimeWeightProducer_cff.py with this sort of thing:

```rules.extend(rulesForLifetimeReweighting['AMSB_chargino_800GeV_20cm_94X'])```

The exact weight values are going to be pushed since 200cm comes from 1000cm, and I'm trying here to reweight way far from 100cm. Then the values I'm storing are:

```
AMSB_chargino_800GeV_20cm_94X
for a chargino with ctau = 100.337, weights_[0] *= 0.000331269 / 0.00366642) = 0.0903521
injected lifetimeWeight_1000024_100cmTo20cm = 0.0903521
default: injected lifetimeWeight = 0.0903521
```
```
AMSB_chargino_800GeV_200cm_94X
for a chargino with ctau = 100.337, weights_[0] *= 0.00302755 / 0.000904533) = 3.34709
injected lifetimeWeight_1000024_1000cmTo200cm = 3.34709
default: injected lifetimeWeight = 3.34709
```
```
AMSB_chargino_800GeV_5cm_94X
for a chargino with ctau = 100.337, weights_[0] *= 3.85364e-10 / 4.38956e-06) = 8.77911e-05
injected lifetimeWeight_1000024_10cmTo5cm = 8.77911e-05
default: injected lifetimeWeight = 8.77911e-05
```
```
AMSB_chargino_800GeV_1000cm_94X
for a chargino with ctau = 100.337, weights_[0] *= 0.00302755 / 0.000904533) = 3.34709
for a chargino with ctau = 100.337, weights_[1] *= 0.00238576 / 0.000904533) = 2.63756
for a chargino with ctau = 100.337, weights_[2] *= 0.00194536 / 0.000904533) = 2.15068
for a chargino with ctau = 100.337, weights_[3] *= 0.00163636 / 0.000904533) = 1.80907
for a chargino with ctau = 100.337, weights_[4] *= 0.00141001 / 0.000904533) = 1.55883
for a chargino with ctau = 100.337, weights_[5] *= 0.0012378 / 0.000904533) = 1.36844
for a chargino with ctau = 100.337, weights_[6] *= 0.00110266 / 0.000904533) = 1.21903
for a chargino with ctau = 100.337, weights_[7] *= 0.000993894 / 0.000904533) = 1.09879
for a chargino with ctau = 100.337, weights_[8] *= 0.000904533 / 0.000904533) = 1
injected lifetimeWeight_1000024_1000cmTo200cm = 3.34709
injected lifetimeWeight_1000024_1000cmTo300cm = 2.63756
injected lifetimeWeight_1000024_1000cmTo400cm = 2.15068
injected lifetimeWeight_1000024_1000cmTo500cm = 1.80907
injected lifetimeWeight_1000024_1000cmTo600cm = 1.55883
injected lifetimeWeight_1000024_1000cmTo700cm = 1.36844
injected lifetimeWeight_1000024_1000cmTo800cm = 1.21903
injected lifetimeWeight_1000024_1000cmTo900cm = 1.09879
injected lifetimeWeight_1000024_1000cmTo1000cm = 1
default: injected lifetimeWeight = 1
```

**Comment by bpf6qc** [comment]
After the last commit, now:

```
rules.extend(rulesForLifetimeReweighting['AMSB_chargino_800GeV_0p9cm_94X'])
...
for a chargino with ctau = 100.337, weights_[0] *= 4.24824e-49 / 2.65587e-44) = 1.59957e-05
injected lifetimeWeight_1000024_1cmTo0p9cm = 1.59957e-05
default: injected lifetimeWeight = 1.59957e-05
```

jalimena:
hi, Displaced Leptons also uses the LifetimeWeightProducer. I believe these changes should be ok/ good for us as well, but I'm hoping that @btcardwell  can also weigh in, since he's used that producer more than I have.

btcardwell:
The LifetimeWeightProducer changes look good to me. I think they'll work fine for Displaced Leptons.

PR #107, opened 2019-02-22
Add new track hit methods

bpf6qc:
Only adds additional methods, doesn't change existing functionality.


jalimena:
Seems alright to me. Would definitely not affect displaced leptons. Is there something you wanted me to test?

PR #108, opened 2019-02-22
invMass operation modified for track with zero-mass assumption

czkaiweb:
In "invMass" operation, if inputCollection is track, track.p is called as track.energy.


aehart:
Looks good @czkaiweb 

Could you just add a comment in the code explaining that for tracks, since they have no measured energy associated with them, we assume they are massless?

czkaiweb:
@aehart  Comment has been added now.

PR #110, opened 2019-02-26
add 2018 muon SFs

jalimena:
starting to add 2018 muon scale factors. however, I can't get ```root -b -q -l recreateLeptonSFs.C``` to work yet - what am I missing? running this command will update our root files with the SFs


jalimena:
ok, with my latest commit it should all be fine, i found and fixed my mistake (I needed to rootify the JSON files, rather than take the root files that the muon POG provides directly, in order to make our format consistent)

bpf6qc:
@jalimena it seems they've very recently added trigger scale factors (gitlab was updated 6 days ago):
https://twiki.cern.ch/twiki/bin/view/CMS/MuonReferenceEffs2018#Results_for_Run2018_data_17Sep20

Analyses will need this because in 2018 the lowest thresholds were IsoMu24, versus IsoMu27 in 2017. The trick with these seems to be separating the trigger scale factors in the middle of 2018A, which the framework is not built to do. But the simplest thing that comes to mind is to store _four_ things (named somehow appropriately?)

BeforeMuonHLTUpdate -- if someone compares 2018A (run < 316361) to MC
AfterMuonHLTUpdate -- if someone compares (run ≥ 316361) to MC
Lumi-weighted average within A -- (ie, lumi(< 316361) / lumi(A) etc) -- if someone compares 2018A to MC
Lumi-weighted average within ABCD -- (ie, lumi(< 316361) / lumi(all 2018)) -- if someone compares 2018ABCD to MC

The fourth one is really what you want, because you can just apply those and not need to worry about dataset names or trying to code an automatic switch between SFs depending on dataset name.

Could you add these please?

bpf6qc:
Beyond adding the trigger SFs these changes are correct to me and I approve of what's been added.

jalimena:
ok, added 2018 muon trigger scale factors. as you suggest, I've added the lumi-weighted SFs. please check.

bpf6qc:
@jalimena I don't see the use case for `AveB` `AveC` or `AveD`, since they are basically the average scale factors if your dataset is B/C/D but with the first 8.9/fb replaced with the first 9.9/fb of A.

If you have a use case I'm not seeing could you mention it? Otherwise I think those 3 (x2 = 6) histograms should be removed.

jalimena:
ok fair enough, i removed them. i'm keeping the text files with the json files and lumi calculations for each 2018 era, though, since they could come in handy in the future (although maybe not in the SF context)

PR #111, opened 2019-02-27
Remove unavailable track methods in 80X

bpf6qc:
HitPattern's TotallyOffOrBad methods had no argument before 94X, so you can't get this value within a inner/middle/outer hit range. This PR just removes these functions from existence before 94X.

### DIFF
--- a/Collections/interface/TrackBase.h
+++ b/Collections/interface/TrackBase.h
@@ -214,6 +214,7 @@ namespace osu
       // a gap of the layer in question; NULL_RETURN hits are not included
       // in this number.
 
+#if CMSSW_VERSION_CODE >= CMSSW_VERSION(9,4,0)
       // expected hits (including inactive modules) differentiated by location in detector
       const unsigned char expectedIncludeInactiveTrackerHits() const;
       const unsigned char expectedIncludeInactivePixelHits() const;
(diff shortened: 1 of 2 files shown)


jalimena:
seems good to me. i'll let @aehart take a look and if he's satisfied, i'd suggest we merge.

aehart:
It's kind of weird (and bad coding in HitPattern) that calling `trackerLayersTotallyOffOrBad()` returns different things in CMSSW_8_X_Y versus CMSSW_9_Z_W. Otherwise I would have suggested simply replacing
```c++
this->hitPattern().trackerLayersTotallyOffOrBad (reco::HitPattern::MISSING_INNER_HITS) +
this->hitPattern().trackerLayersTotallyOffOrBad (reco::HitPattern::TRACK_HITS) +
this->hitPattern().trackerLayersTotallyOffOrBad (reco::HitPattern::MISSING_OUTER_HITS))
```
with
```c++
this->hitPattern().trackerLayersTotallyOffOrBad ()
```
so the same code worked in 8 and 9. But the latter would only give you the totally off or bad _middle_ layers in 9.

Anyway, assuming we don't actually need these methods in ≥9, it the changes look fine, and I'll merge.

Issue #87, opened 2018-11-27
update to electron Fall17 VIDs V2


jalimena:
When we are ready to re-intuplize, we should update to the Fall17 V2 version of the electron VID. See https://github.com/OSU-CMS/OSUT3Analysis/pull/86

Basically, the "False" needs to be removed, and the updated compiling procedure needs to be implemented for CMSSW_9_4_9 and greater.

bpf6qc:
The VID implementation should be checked as well given changes in:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/EgammaMiniAODV2#2017_MiniAOD_V2

Also:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/EgammaMiniAODV2#Accessing_ID_result
is potentially important because the ValueMap associating electrons to VID values was very difficult to reproduce in multiple electron collections, this may make things much easier. (e.g. this here is a note to myself for when I look at doing this)

jalimena:
see also #113

jalimena:
ok so we can close this issue now, right? given #112  and #113 ?

bpf6qc:
@jalimena This issue is somewhat (?) still relevant I suppose, that users need to know this flag exists to control V2 IDs in 94X. But #112 and #113 do have the decision that we'll keep things the way they are, so it's not exactly an "issue" in github parlance anymore. So I would be in favor of closing this.

jalimena:
Yeah, I think it is no longer an issue we need to take care of, but rather an option we can do now. So I'm going to close this issue, and put a note about this option of using V2 in 94X on our analysis recipe twiki. I suggest you do the same.

PR #112, opened 2019-03-04
Add 2017v2, 2018 electron SFs

bpf6qc:
https://twiki.cern.ch/twiki/bin/view/CMS/EgammaIDRecipesRun2

Electron SFs are always pretty straightforward. There's a V2 that appeared for 2017 since last we looked, and I've added them with a version/wp suffix "V2". The veto ID is missing from 2017v2 and 2018 but there's not much we can do about that.

Lastly the changes to recreateLeptonSFs.C compete with PR #110 so we will need to merge them.


jalimena:
hi @bpf6qc  - what you've implemented here looks good to me. however, does this PR deal with all the issues we mentioned in https://github.com/OSU-CMS/OSUT3Analysis/issues/87 ? Probably not, right?

bpf6qc:
@jalimena no it doesn't, although I'm looking at that now. #87 is just as important as this PR so perhaps I should fold it into this one.

jalimena:
so now that the muon SF PR is merged, do you want to rebase and then we merge this one? or we wait until you add the info in #87?

bpf6qc:
I've rebased this PR with PR #110, and instead of waiting to merge these I've created a separate PR #113 .

bpf6qc:
@jalimena yes these two PRs are now pretty disjoint in their effects. I'll merge this now, thanks!

PR #113, opened 2019-03-06
Add2018ele trigger match

bpf6qc:
A few things:

1) Renamed `passesTightID_noIsolatedion` to be very clear about what it is -- the only group using this should be Disappearing Tracks if they need to go back to 2015-6. Otherwise this ID is quite deprecated, and the VIDs provide real IDs.
2) Adds a trigger match method for Ele32 which is present in 2018 data.
3) Fixes a mistake in the track class for matching to PF muons
4) Clarifies a few CMSSW version checks

Overall I've decided that for Issue #87, for now, I won't switch to this new `electron.electronID()` method. While it would make a few things simpler, it's turned out to be a bit uncertain which CMSSW releases will have which versions available, and if this will be compatible for all analyses. I think we have something that works now and I don't want to start breaking it too much.

This PR needs a little more checking, I'll comment when I'm done.

### DIFF
--- a/Collections/interface/Electron.h
+++ b/Collections/interface/Electron.h
@@ -35,7 +35,7 @@ namespace osu
         const int missingMiddleHitsFromTrackerLayersWithoutMeasurements () const;
         const int missingOuterHitsFromTrackerLayersWithoutMeasurements () const;
 
-        const bool passesTightID_noIsolation () const;
+        const bool passesTightID_noIsolation_LegacySpring15 () const;
 
         const bool passesVID_vetoID () const;
         const bool passesVID_looseID () const;
@@ -44,6 +44,7 @@ namespace osu
 
         const bool match_HLT_Ele25_eta2p1_WPTight_Gsf_v () const;
         const bool match_HLT_Ele22_eta2p1_WPLoose_Gsf_v () const;
+        const bool match_HLT_Ele32_WPTight_Gsf_v () const;
         const bool match_HLT_Ele35_WPTight_Gsf_v () const;
 
         void set_rho (float value) { rho_  = value; }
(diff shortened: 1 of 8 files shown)


bpf6qc:
I'm happy with this PR enough now to request reviews.

jalimena:
Also, on your point 1 above: what do we do now if we need the VID without isolation?

bpf6qc:
@jalimena in the case of disappearing tracks, we've used the V1 ID's in 94X so at this point we probably shouldn't redo everything. The only thing I can think of beyond just having users change the `False` in their code recipes, is to provide another set of functions to get the V1/V2 results. Then you could switch which function you cut on in your analysis code.

And the VID without isolation, this actually is another point in favor of keeping the on-the-fly ValueMap method, because `electron.electronID("id-name")` will not allow you to remove cuts. The method to remove a cut is still to use something like:
```removeVIDCut(process, 'cutBasedElectronID-Summer16-80X-V1-tight', 'GsfEleEffAreaPFIsoCut')```
to expressly remove that section of the ID cuts. You'd need to examine the ID module and find the name of the cut you want to remove, if it's not `GsfEleEffAreaPFIsoCut`.

One last question then, would you prefer we kept the `passesTightID_noIsolatedion` function or do you want to use `removeVIDCut`? I could add it back with a better name, say `passesTightID_noIsolatedion_Fall17V1`.

jalimena:
Ok thanks for explaining, sorry I forgot about removeVIDCut. We are in fact using it, I just forgot.

So then, the PR as is looks good to me. If you do any sort of quick runtime test, I think we can merge it.

bpf6qc:
@jalimena I've compiled this in 80X, 94X, and 10_2_X successfully. The only "real" change here is adding the HLT matching for Ele32 in 2018 data, and only in 2018 data, and it does indeed work. In 2018 the V2 IDs and Ele32 are available and I see electrons passing my selection.

PR #115, opened 2019-03-10
Add siblings as secondary files according to sibling_datasets dictionary.

aehart:
The intended use case here is that you have ntuples produced from a higher data tier, e.g., AOD, that contain a special collection not found in MINIAOD, e.g., tracks. And you want to run over these ntuples, using the standard MINIAOD collections for all of the other object types.

With the current changes, you can define a dictionary in your local config file called `sibling_datasets`, e.g.:
```python
sibling_datasets = {
    "MET_2018A" : "/MET/Run2018A-17Sep2018-v1/MINIAOD",
    "MET_2018B" : "/MET/Run2018B-17Sep2018-v1/MINIAOD",
    …
}
```
Then the config files that `osusub.py` generates will add the necessary files from the values of this dictionary to the secondary files in `process.source`.

If this dictionary is not defined, then the behavior is unchanged.

### DIFF
--- a/DBTools/python/osusub_cfg.py
+++ b/DBTools/python/osusub_cfg.py
@@ -26,3 +26,51 @@ elif len (sys.argv) == 5 and sys.argv[2] == "True":
 else:
   batchMode = False
 
+# Function for getting siblings of a given file from a given dataset
+def getSiblings (fileName, dataset):
+  try:
+    from dbs.apis.dbsClient import DbsApi
+    from CRABClient.ClientUtilities import DBSURLS
+  except ImportError:
+    print "getSiblings() relies on CRAB. Please set up the environment for CRAB before using."
+    sys.exit (1)
+
+  dbsurl_global = DBSURLS["reader"].get ("global", "global")
+  dbsurl_phys03 = DBSURLS["reader"].get ("phys03", "phys03")
+  dbs3api_phys03 = DbsApi (url = dbsurl_phys03)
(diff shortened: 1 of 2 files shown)


jalimena:
seems reasonable. how have you tested this?

aehart:
I've run a test job over one of our 2018 ntuples, which have only the track collection that we use saved. It is able to find the correct sibling files, according to the dictionary that I added in DisappTrks, and it successfully processes events.

When I comment out the sibling_datasets dictionary, it generates config files identical to what is generated now, before this PR is merged. And when I submit jobs without a grid proxy, e.g., as one would do when running over a local skim, it fails to contact DAS and does not add the sibling files, as it should in that case.

bpf6qc:
I'm actually having trouble getting these jobs to run. I see the error message:
```No valid grid proxy. Not adding sibling files.```
which is not true, voms-proxy-info shows I should be okay. Somehow the `try` sections added are failing for `getSiblings` but not in the CRAB import stuff.

What could I be missing? I think there must be some failure mode with no clear warning message.

jalimena:
So, is this PR ready? Has it been tested well by Andrew and Brian? It would be good to merge it soon because NOW the rereco json file has been announced:
https://hypernews.cern.ch/HyperNews/CMS/get/physics-validation/3326.html

czkaiweb:
I have  tested --forceRutgerMode with TauSkim and it works.
Maybe it  is time to merge this pull request.

PR #116, opened 2019-03-11
Fix source and destination cTaus that had accidentally been switched

btcardwell:
The DisplacedSUSY source and destination cTau definitions were accidentally switched in a previous PR. This PR swaps them back to their proper places.


jalimena:
works for me. merging.

PR #118, opened 2019-03-25
Extended siblings

aehart:
The main change in this PR is in osusub_cfg.py, where the `getSiblings` function has been extended. Before, it would get a list of siblings from the prod/global DBS instance, assuming the given file was in prod/phys03. Now, the reverse case is handled, where the list of siblings comes from the prod/phys03 instance, assuming the given file is in prod/global. The case is determined by whether the target dataset ends in "/USER".

A few other minor changes have been made to osusub.py:
- When in "Rutgers mode", the proxy file is copied to /tmp/ on the worker nodes, instead of to the same directory that it lives in on the interactive nodes. This is needed on the LPC because the home directories, where the proxy file now lives on the interactive nodes, is not mounted on the worker nodes.
- In the config_cfg.py file that is generated by osusub.py, the list of siblings is only gotten when running in batch mode. This prevents contacting DBS when this file is imported during the submission process.
- No redirector is prepended to each file name if the user does not use the --redirector option.

### DIFF
--- a/DBTools/python/osusub_cfg.py
+++ b/DBTools/python/osusub_cfg.py
@@ -101,7 +101,7 @@ def getSiblings (fileName, dataset):
     # put the files of the target dataset in another set
     dataset = dbs3api_phys03.listFiles (dataset = dataset)
     for f in dataset:
-      miniaodSuperset.add (f["logical_file_name"])
+      miniaodSubset.add (f["logical_file_name"])
 
   # return the intersection of the two sets
   return list (miniaodSubset.intersection (miniaod))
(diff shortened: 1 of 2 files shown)


jalimena:
i think it's best if @bpf6qc and/or @czkaiweb  review and test this PR

bpf6qc:
@czkaiweb and @aehart you're both already running with this PR successfully, is that right? If so @czkaiweb should comment and merge, otherwise comment about remaining issues.

czkaiweb:
Yes，I have tested this PR and we can merge it.

PR #119, opened 2019-03-26
L1 ECAL prefiring weight producer

bpf6qc:
See:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/ExoPreapprovalChecklist
https://twiki.cern.ch/twiki/bin/viewauth/CMS/L1ECALPrefiringWeightRecipe

Creates an eventvariable producer for the L1 ECAL prefiring issue in 2016-7 data, to simulate the effect in MC. This code is used by:
```
variableProducers = []
...
variableProducers.extend(l1PrefiringWeightVariableProducers)
```
One will need to check out the code in 2017 (2016) releases with:
```
git cms-merge-topic lathomas:L1Prefiring_9_4_9 (8_0_32)
```
and `setupFramework.py` will give a reminder to do so in these releases.

Doing this, additional eventvariables are then available:
`L1ECALPrefiringWeight(Up/Down)` -- the event weight and its up/down fluctuations.
`hasPrefiredJets` -- a boolean, true if there are any jets with pt > 100 GeV in 2.25 < |eta| < 3.0.

The user can apply the event weight as appropriate. It is currently the responsibility of each analysis' code to set the era for weights, which by default are set to:
```DataEra = cms.string("2017BtoF")```

These values are set to 1.0 and false if the event isn't MC.

### CODE REVIEW

File: AnaTools/plugins/L1PrefiringWeightProducer.h [hunk 1]
```diff
@@ -0,0 +1,22 @@
+#ifndef PUSCALING_FACTOR_PRODUCER
+#define PUSCALING_FACTOR_PRODUCER
```
aehart: Can you change the macro name here?

aehart:
Remembering to set the data era seems like something that will be really easy to forget. Could we at least set it automatically when in batch mode, using the dataset label in osusub_cfg.py?

I just don't want to accidentally apply these scale factors to signal MC for 2018, for example.

AnaTools/plugins/L1PrefiringWeightProducer.h
bpf6qc: Changed, thanks for catching this.

bpf6qc:
@aehart Are we sure we want to control the behavior of this within something like `L1PrefiringWeightProducer_cff.py`? For signal you might have dataset labels with suffixes like `94X`, but typically they would be `TTJets` or something with no reference to what year they might belong. 

You could instead key off of `CMSSW_VERSION`, but you'd still need the analysis-specific code to avoid adding the producer in 2018 conditions -- within `L1PrefiringWeightProducer_cff.py` the best you could do is give a garbage era that would produce a failure if someone is still adding the producer. Also there's the ultra-legacy re-reco of 2016 data that will happen in 10_2_X; no one is in this situation yet but it may happen.

My thinking was, in the example of DisappearingTrks, to configure the era within customize.py. That way the analyses would control this and prevent themselves from getting it wrong.

jalimena:
Looks good. When @aehart and @bpf6qc  agree on the issue above and the version is committed, I would like to test this from the displaced leptons point of view before merging.

bpf6qc:
@aehart I've added a check on the `DataEra` copied from the CMSSW producer, and set the default value to a bad value. The CMSSW producer terribly does not care about this, and would fail silently producing all weights as `1.0` without saying anything.

This is a step short from manually checking the values of the weights, since at that point we'd be basically reproducing the code from the CMSSW module. This latest commit requires the user to change both strings or face a fatal error:

```
if hasattr (process, "prefiringweight"):
            process.prefiringweight.DataEra = cms.string("2017BtoF")
            process.L1PrefiringWeightProducer.DataEra = cms.string("2017BtoF")
```

Now technically if the user tries _really_ hard they could just change one and not the other and get all 1's...but this is at least some check.

aehart:
@bpf6qc Thanks for adding that check. Looks good to me.

jalimena:
ok, it looks like you two have converged. So maybe @bpf6qc  can implement this for disappearing tracks, test it, and then send that to me as an example? Then i can implement it for displaced leptons and test it as well. Thanks!

bpf6qc:
@jalimena for an example, we've implemented this now and it works fine.

Including the variable producer:
https://github.com/OSU-CMS/DisappTrks/blob/master/StandardAnalysis/python/protoConfig_cfg.py#L244-L246

Customizing the DataEra (it's set to junk by default):
https://github.com/OSU-CMS/DisappTrks/blob/master/StandardAnalysis/python/customize.py#L140-L142

Including the weight in the Plotter:
https://github.com/OSU-CMS/DisappTrks/blob/master/StandardAnalysis/python/EventWeights.py#L27-L34
(also at the bottom of that module is the Up/Down fluctuations for systematics)

bpf6qc:
@jalimena oh and

https://github.com/OSU-CMS/DisappTrks/blob/master/StandardAnalysis/python/Cuts.py#L162-L166

To study if this issue affects your analysis. You'd need to include this producer and customize DataSet and `git cms-merge-topic lathomas:L1Prefiring_9_4_9 (8_0_32)` and all that too, unless I make further changes to separate the bool from the weights in the producer.

jalimena:
pointing @marthaEstefany to this PR, in case she has time to test it for displaced leptons

marthaEstefany:
In order to test that this worked in DisplacedSUSY, I made a new testing area using CMSSW_8_0_32 and ran a few events interactively. It works interactively. If we find that we are affected by this weight and need to use it, we will have to move from CMSSW_8_0_31 to CMSSW_8_0_32. I am currently running our signal region without the veto cut and with the veto cut to see whether or not we are affected by more than 1%.

jalimena:
Hi @marthaEstefany , thanks for testing this. Which sample(s) are you testing it on? (Don't look at data in our signal region!) And you're running over a prelection skim, right?

bpf6qc:
@jalimena @marthaEstefany keep in mind testing this PR is a separate issue from using it to evaluate your analysis :)

Are there comments on how to improve these changes?

jalimena:
yeah fair point, sorry about that, @bpf6qc . You said it works fine for disappearing tracks, so i'll merge.

PR #121, opened 2019-04-01
Add AMSB reweighting rules from 10cm

bpf6qc:
Extends the rules for 10cm AMSB samples. The only thing this changes is that additional eventvariables are created when running LifetimeWeightProducer over these samples, reweighting 10cm all the way down to 0.1cm.

Tested by manually calculating the weight for 10cm --> 0.2 (and 0.9) cm using the chargino lifetimes:
```
Total weight from eventvariables: 0.0306425 (2cm), 2.55763e-31 (0p2cm)
Total weight from math:           0.0306425 (2cm), 2.55763e-31 (0p2cm)
```

So the eventvariables are produced and they have the right values.

### DIFF
--- a/Configuration/python/configurationOptions.py
+++ b/Configuration/python/configurationOptions.py
@@ -9673,6 +9673,8 @@ for dataset0 in nJobs:
     pdgIdsForISRReweighting[dataset] = [1000022, 1000024]
 
   destinationCTaus = [ctau0 * 0.1 * i for i in range(2, 11)]
+  if ctau0 == 10:
+    destinationCTaus.extend([round(ctau0 * 0.01 * i, 2) for i in range(2, 11)])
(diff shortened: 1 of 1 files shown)


jalimena:
In principle this seems fine to me, but I'm having trouble tracking the changes. When I click "view this file", it seems that many more changes should have already been made at the end of configurationOptions.py? For example, there is no mention in the master version of configurationOptions.py of "pdgIdsForISRReweighting". What am I missing? Did an earlier PR not get merged correctly?

bpf6qc:
@jalimena I also see that in the file display, however if I view the 'raw' text I see the file just fine. It's probably just a limitation of the site or it's messing up because this file is nearly 10k lines long.

aehart:
I think GitHub is telling us we need to refactor configurationOptions.py by truncating it at line 9564 in the non-raw view.

Anyway, this change looks good to me, so I'll merge it.

PR #122, opened 2019-04-09
passBadCalibFilterUpdate

czkaiweb:
passBadCalibFilterUpdate is recommended in 
https://twiki.cern.ch/twiki/bin/viewauth/CMS/MissingETOptionalFiltersRun2#2018_data
Changes are adopted in METfilter mainly for 2018 datasets.
Please check the change..

### DIFF
--- a/Collections/interface/Met.h
+++ b/Collections/interface/Met.h
@@ -58,9 +58,11 @@ namespace osu
 
         void setBadChargedCandidateFilter (const bool);
         void setBadPFMuonFilter (const bool);
+        void setPassecalBadCalibFilterUpdate (const bool);
 
         const bool badChargedCandidateFilter () const { return badChargedCandidateFilter_; }
         const bool badPFMuonFilter () const { return badPFMuonFilter_; }
+        const bool passecalBadCalibFilterUpdate () const { return passecalBadCalibFilterUpdate_; }
 
       private:
         double noMuPt_;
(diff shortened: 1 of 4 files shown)


aehart:
Thanks for implementing this @czkaiweb 

Looks good. My only suggestion would be to rename the flag from `passecalBadCalibFilterUpdate` to `ecalBadCalibFilter` to be more in line with the other flags defined in osu::Met.

bpf6qc:
@aehart I initially suggested he keep the `updated` word on this new flag, because there will continue to be the old flag in MINIAOD. I was thinking that analyses might want to keep both around in case the MET scanners change their minds yet again.

aehart:
@bpf6qc Fair point. I was just thinking that there's nothing specific to the update in the code here. It's up to the individual analyses to rerun the filter with an updated list of bad ECAL channels. And the old flag is still available via the `metFilters` parameter to `CutCalculator`.

In any case, I don't feel strongly about this, so whatever you decide is fine by me.

jalimena:
All looks fine to me. No strong opinion on the name here. Displaced leptons never looks at MET, so we are not affected.

bpf6qc:
I don't see any movement on this PR. @jalimena, @aehart if you're okay with this I suggest we merge this.

jalimena:
so we agreed to keep the name as it is? if so then feel free to merge

PR #123, opened 2019-04-10
Edit lepton pT hists

btcardwell:
DisplacedSUSY needs leading and sub-leading electron pT histograms for our new background estimation technique, and I'd like to look at the lepton pT spectra with finer bins. Changes work when tested interactively.


bpf6qc:
This is fine to me @btcardwell, @jalimena; disappearing tracks won't need these plots but I won't turn down more information.

PR #124, opened 2019-04-13
Update IsoMu24/27 for 2018

czkaiweb:
The IsoMu24 trigger filter in 2018 has different names. The tags are set with CollectionProducer_cff.cfg.
Further check for HLT path should be done before merging.


jalimena:
Given what I know about the muon HLT changes in 2018, these name changes make sense.  You might also check if the final filter for HLT_IsoMu20 has changed. See https://twiki.cern.ch/twiki/bin/view/CMS/MuonHLT2018#Muon_HLT_Task_Force for more details about the change.

Configuration/python/CollectionProducer_cff.py
```diff
@@ -160,15 +160,27 @@ class CollectionProducer:
...
     d0SmearingWidth = cms.double (-1.0),
+    hltMatchingInfo = cms.VPSet (
+        cms.PSet(name = cms.string("HLT_IsoMu20_v"),   collection = cms.string("hltL3MuonCandidates::HLT"),     filter = cms.string("hltL3crIsoL1sMu16L1f0L2f10QL3f20QL3trkIsoFiltered0p09")),
+        cms.PSet(name = cms.string("HLT_IsoTkMu20_v"), collection = cms.string("hltHighPtTkMuonCands::HLT"),    filter = cms.string("hltL3fL1sMu16L1f0Tkf20QL3trkIsoFiltered0p09")),
+        cms.PSet(name = cms.string("HLT_IsoMu24_v"),   collection = cms.string("hltL3MuonCandidates::HLT"),     filter = cms.string("hltL3crIsoL1sMu22L1f0L2f10QL3f24QL3trkIsoFiltered0p09")),
+        cms.PSet(name = cms.string("HLT_IsoTkMu24_v"), collection = cms.string("hltHighPtTkMuonCands::HLT"),    filter = cms.string("hltL3fL1sMu22L1f0Tkf24QL3trkIsoFiltered0p09")),
+        cms.PSet(name = cms.string("HLT_IsoMu27_v"),   collection = cms.string("hltIterL3MuonCandidates::HLT"), filter = cms.string("hltL3crIsoL1sMu22Or25L1f0L2f10QL3f27QL3trkIsoFiltered0p07")),
+    ),
 )
+if os.environ["CMSSW_VERSION"].startswith ("CMSSW_10_2_"):
```
jalimena: It looks like just IsoMu24 is now updated to the right filter in 10_2_X? What about IsoMu27?

bpf6qc: I haven't been able to dump the HLT menu for 2018 to check for other changes, but I'm thinking this might be okay because IsoMu24 is the recommended and un-prescaled path in 2018. Is there a use case for the other paths? We can still check this though.

bpf6qc:
@jalimena after getting the menu it looks like IsoMu20 was changed in 2018, but I see the same filter for IsoMu27 as was in 2017.

jalimena:
ok, all seems good, agreed. I think we're ready to merge now?

PR #126, opened 2019-04-30
add stop -> ld besides stop -> lb for displaced SUSY signal

jalimena:


btcardwell:
This all looks good to me

jalimena:
can we merge or @bpf6qc , would you like to take a look? should be relatively straightforward and only should affect displaced leptons.

PR #127, opened 2019-05-09
Add numTrueInteractions to eventvariables

bpf6qc:
Makes `numTrueInteractions` (BX0->getTrueNumInteractions()) available in eventvariables. Set to a large negative number if the event isn't MC or if not using MINIAOD.


jalimena:
seems fine to me. i'll wait for @aehart before merging.

PR #128, opened 2019-05-14
Add soft muon

marthaEstefany:
I add a method to get the soft muon id. I also added a low mass mu-mu invariant mass histogram definition.


bpf6qc:
The added ID function looks good to me and Estefany showed me that muons pass her selection just fine. I asked her to remove the extra dimuon mass plot (but leave the isSoftMuonID plot) because she will need a large number of bins for a J/Psi analysis, much more than 10, and no other analysis will need such a large histogram in their output files.

jalimena:
This seems fine indeed, and I agree it's good to keep the relevant invariant mass histogram in the DisplacedSUSY package (@marthaEstefany , please put that histogram in DisplacedSUSY and push the change).

One question I have is whether we should define the soft muon wrt the beam spot, because we define it like that for tight muons elsewhere in the analysis, rather than wrt to primary vertex. However, we can certainly have this function in OSUT3Analysis, so i'll merge.

PR #130, opened 2019-05-22
fix for 3D plots

jalimena:
We need to change a 1 to a 2, so that 3D plots with different indices get filled.

See https://milliqanelog.asc.ohio-state.edu:8080/OSUT3Analysis/140 and
https://milliqanelog.asc.ohio-state.edu:8080/OSUT3Analysis/141


bpf6qc:
So the bug is if you provide an `indexZ` value for a 3d histogram, then you're mixing iterators which are offset in memory and it always continues. Neat.

Good catch and fix.

jalimena:
exactly. thanks!

PR #131, opened 2019-06-08
Save decay length, beta, and gamma along with cTau for plotting

bpf6qc:
Adds eventvariables like `decayLength_1000024_0` for the decay length, beta, and gamma of particles used in the lifetime reweighting, to go along with whatever `cTau_1000024_1` (etc) variables you're already seeing.

In a DisappTrks repo version of `getEndVertex`, we require the daughter ID to be the neutralino, but this is of course too specific for the framework. The vertices of all daughters should really be the same so I've used that as a shortcut.

### DIFF
--- a/AnaTools/plugins/LifetimeWeightProducer.cc
+++ b/AnaTools/plugins/LifetimeWeightProducer.cc
@@ -75,7 +75,10 @@ LifetimeWeightProducer::AddVariables(const edm::Event &event) {
 
   for(unsigned int iRule = 0; iRule < dstCTau_.size(); iRule++) {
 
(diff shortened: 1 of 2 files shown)


bpf6qc:
Running this over some samples of AMSB and plotting the tree value:

```eventvariable_decayLength_1000024_0 / eventvariable_gamma_1000024_0 / eventvariable_beta_1000024_0 / eventvariable_cTau_1000024_0```

The values are all identically `1`. @jalimena something else has come up however, I will be adding one more variable that can go with this PR.

bpf6qc:
@jalimena @aehart we actually don't need this other thing, so this PR is complete with the check above.

PR #135, opened 2019-07-10
Update 2018 fiducial maps with more data

bpf6qc:
Only missing EGamma_2018D now. The `beforeVeto` and `afterVeto` histograms for 2018-inclusive are just ABC for EGamma, so users will need to keep that in mind.

Screen Shot 2019-07-10 at 11 40 01 AM


czkaiweb:
Looks fine

PR #136, opened 2019-07-12
Match to isolated tracks

bpf6qc:
Adds a match to the nearest pat::IsolatedTrack (within dR <= 0.2) if the data format has them. None of this affects anyone who does not `#define DISAPP_TRKS`.

Example in AMSB:
Screen Shot 2019-07-12 at 8 47 02 AM


jalimena:
looks fine from my point of view, but i have not explicitly tested it. probably best if @aehart  and/or @czkaiweb  takes a look?

PR #145, opened 2019-11-20
add cosine of 3D angle between 2 objects

jalimena:
Adds a method to find the cosine of the 3D angle between 2 objects. 

Can someone check if what I'm doing for the bool at the end of L838-843 makes sense? I think you put "false" until you want to change the object iteration - is that right?

### DIFF
--- a/AnaTools/src/ValueLookupTree.cc
+++ b/AnaTools/src/ValueLookupTree.cc
@@ -509,7 +509,7 @@ ValueLookupTree::insert_ (const string &cut, Node * const parent) const
                                                   "ceil", "floor", "fmod", "trunc", "round", "rint", "nearbyint", "remainder", "abs", "fabs",
                                                   "copysign", "nextafter",
                                                   "fdim", "fmax", "fmin", "max", "min"}) ||
-        insertUnaryPrefixOperator  (cut,  tree,  {"deltaPhi", "dPhi", "normalizedPhi", "compositePhi", "deltaR", "invMass", "number", "transMass", "pT"}) ||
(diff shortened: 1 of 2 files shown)


jalimena:
some evidence that the variable and histogram are working as expected: https://milliqanelog.asc.ohio-state.edu:8080/DisplacedSUSY/1410

bpf6qc:
@jalimena your changes are good, however in checking your question about the `shouldIterate` flag I see problems in two other places that you can fix.

L 775-778 in compositePhi, this must be changed to:
```
          px0 = valueLookup (boost::get<string> (operands.at (0)) + "s", objs, "px");
          py0 = valueLookup (boost::get<string> (operands.at (0)) + "s", objs, "py", false);
          px1 = valueLookup (boost::get<string> (operands.at (1)) + "s", objs, "px");
          py1 = valueLookup (boost::get<string> (operands.at (1)) + "s", objs, "py", false);
```

L 820-823 in transMass, this must be changed to:
```
          double pt0 = valueLookup (boost::get<string> (operands.at (0)) + "s", objs, "pt"),
                      phi0 = valueLookup (boost::get<string> (operands.at (0)) + "s", objs, "phi", false),
                      pt1 = valueLookup (boost::get<string> (operands.at (1)) + "s", objs, "pt"),
                      phi1 = valueLookup (boost::get<string> (operands.at (1)) + "s", objs, "phi", false);

         double dPhi = deltaPhi (phi0, phi1);
```
I'll let you touch up the style/indentation as you like but the flag at the end is wrong for these functions.

These functions have worked in the case where one of the collections has exactly one element, for example `transMass(muons, mets)` will calculate things correctly, but will segfault if you use `transMass(muons, muons)`. We have just been lucky so far that no one has tried to do this!

PR #148, opened 2019-12-10
Add stop masses

marthaEstefany:
adding stop masses 150 and 175 GeV for lifetimes 1mm and 1000mm


jalimena:
if we can squash all these different merges, however, that would be nice. @marthaEstefany , please update this PR with the merges squashed, or let us know that it's too hard to do quickly and we should just forget about it.

marthaEstefany:
OK. I think I did the squashing thing.

jalimena:
excellent, thanks!

PR #149, opened 2020-02-06
Add photon VID and basic photon plots

btcardwell:
This PR adds photon VID following the example of the electron VID implementation. It works when tested in CMSSW 10_2_12. I'll test it in 8_0 and 9_4 as well, but I figured I'd make the PR now so people could take a look.

### DIFF
--- a/Configuration/python/CollectionProducer_cff.py
+++ b/Configuration/python/CollectionProducer_cff.py
@@ -205,7 +205,7 @@ if os.environ["CMSSW_VERSION"].startswith ("CMSSW_9_4"):
     collectionProducer.photons.effAreasChargedHadronPayload = cms.FileInPath ("RecoEgamma/PhotonIdentification/data/Fall17/effAreaPhotons_cone03_pfChargedHadrons_90percentBased.txt")
     collectionProducer.photons.effAreasNeutralHadronPayload = cms.FileInPath ("RecoEgamma/PhotonIdentification/data/Fall17/effAreaPhotons_cone03_pfNeutralHadrons_90percentBased.txt")
     collectionProducer.photons.effAreasPhotonPayload        = cms.FileInPath ("RecoEgamma/PhotonIdentification/data/Fall17/effAreaPhotons_cone03_pfPhotons_90percentBased.txt")
-    if (False and int(os.environ["CMSSW_VERSION"].split("_")[3]) >=9):
+    if (False and int(os.environ["CMSSW_VERSION"].split("_")[3]) >= 13):
         collectionProducer.photons.vidLooseIdMap  = cms.InputTag ("egmPhotonIDs:cutBasedPhotonID-Fall17-94X-V2-loose")
         collectionProducer.photons.vidMediumIdMap = cms.InputTag ("egmPhotonIDs:cutBasedPhotonID-Fall17-94X-V2-medium")
         collectionProducer.photons.vidTightIdMap  = cms.InputTag ("egmPhotonIDs:cutBasedPhotonID-Fall17-94X-V2-tight")
--- a/Configuration/python/CollectionProducer_cff.py
+++ b/Configuration/python/CollectionProducer_cff.py
@@ -198,7 +198,22 @@ collectionProducer.photons = cms.EDProducer ("OSUPhotonProducer",
     effAreasPhotonPayload        = cms.FileInPath ("RecoEgamma/PhotonIdentification/data/Spring16/effAreaPhotons_cone03_pfPhotons_90percentBased.txt"),
 )
 
-if os.environ["CMSSW_VERSION"].startswith ("CMSSW_9_4") or os.environ["CMSSW_VERSION"].startswith("CMSSW_10_2"):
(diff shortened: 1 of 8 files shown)


jalimena:
seems reasonable to me

btcardwell:
Alright, I've now tested in CMSSW_8_0, 9_4, and 10_2. Everything appears to be working, so I believe this PR is ready to be reviewed and merged. Note that one must follow the instructions in [0] to use photon VID in 8_0.

[0] https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedPhotonIdentificationRun2#Recipe_for_regular_users_for_8_0

Configuration/python/processingUtilities.py
```diff
@@ -979,6 +990,34 @@ def customizeMINIAODElectronVID(process, collections, usedCollections):
 
     return process
 
+def customizeMINIAODPhotonVID(process, collections, usedCollections):
+    if not hasattr (collections, "photons"):
```
bpf6qc: I think this should be
`if not hasattr (collections, "photons") or "photons" not in usedCollections:`
and similar for `customizeMINIAODElectronVID`.

Everyone has photons/electrons defined in `collections`, but I'm pretty sure this check should just skip adding CPU-costly VID modules when you're not cutting on or plotting these things.

I'm ~pretty sure but it's worth a test.

Configuration/python/processingUtilities.py
```diff
@@ -979,6 +990,34 @@ def customizeMINIAODElectronVID(process, collections, usedCollections):
...
+        return process
+
+    from PhysicsTools.SelectorUtils.tools.vid_id_tools import DataFormat,switchOnVIDPhotonIdProducer,setupAllVIDIdsInModule,setupVIDPhotonSelection
+    switchOnVIDPhotonIdProducer(process, DataFormat.MiniAOD)
+
+    my_id_modules = ['RecoEgamma.PhotonIdentification.Identification.cutBasedPhotonID_Spring16_V2p2_cff']
+
+    if os.environ["CMSSW_VERSION"].startswith ("CMSSW_9_4_"):
+        my_id_modules = ['RecoEgamma.PhotonIdentification.Identification.cutBasedPhotonID_Fall17_94X_V1_cff']
+        if int(os.environ["CMSSW_VERSION"].split("_")[3]) >=9:
```
bpf6qc: For 9_4_X, I can only find `RecoEgamma/PhotonIdentification/python/Identification/cutBasedPhotonID_Fall17_94X_V2_cff.py`
with X >= 13. I had to look in CVMFS to discover this, because the github repo only shows 9_4_X all together.

I can't find any recipes to provide this payload in earlier versions, so I would recommend two things:
    1. Change this check to `>= 13`, and do the same in CollectionProducer_cff
    2. Add a warning to setupFramework.py that if 9_4 and yet < 13, the photon VID version is only V1, and the user should consider moving to a newer release

Configuration/python/CollectionProducer_cff.py
```diff
@@ -190,7 +190,37 @@ class CollectionProducer:
...
 )
+
+if os.environ["CMSSW_VERSION"].startswith ("CMSSW_9_4"):
+    collectionProducer.photons.vidLooseIdMap  = cms.InputTag ("egmPhotonIDs:cutBasedPhotonID-Fall17-94X-V1-loose")
+    collectionProducer.photons.vidMediumIdMap = cms.InputTag ("egmPhotonIDs:cutBasedPhotonID-Fall17-94X-V1-medium")
+    collectionProducer.photons.vidTightIdMap  = cms.InputTag ("egmPhotonIDs:cutBasedPhotonID-Fall17-94X-V1-tight")
+    collectionProducer.photons.effAreasChargedHadronPayload = cms.FileInPath ("RecoEgamma/PhotonIdentification/data/Fall17/effAreaPhotons_cone03_pfChargedHadrons_90percentBased.txt")
+    collectionProducer.photons.effAreasNeutralHadronPayload = cms.FileInPath ("RecoEgamma/PhotonIdentification/data/Fall17/effAreaPhotons_cone03_pfNeutralHadrons_90percentBased.txt")
+    collectionProducer.photons.effAreasPhotonPayload        = cms.FileInPath ("RecoEgamma/PhotonIdentification/data/Fall17/effAreaPhotons_cone03_pfPhotons_90percentBased.txt")
+    if (False and int(os.environ["CMSSW_VERSION"].split("_")[3]) >=9):
```
bpf6qc: See comment below about `>=9` --> `>=13`

Configuration/python/processingUtilities.py
bpf6qc: Also for CMSSW 8_0_X, this twiki:

https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedPhotonIdentificationRun2#Recipe_for_regular_users_for_8_0

says a recipe is needed. In 8_0_X please also add a check to `setupFramework.py` similar to the other warnings at the bottom.

btcardwell:
I implemented all the suggested changes and all seems good after another round of tests. I think this PR is ready to merge if everyone is on board.

PR #150, opened 2020-02-07
Gen match dr

jalimena:
add the dR between a particle and its gen match
then add these dR histograms for electrons and muons

### DIFF
--- a/Configuration/python/histogramDefinitions.py
+++ b/Configuration/python/histogramDefinitions.py
@@ -533,7 +533,7 @@ ElectronHistograms = cms.PSet(
             title = cms.string(";#DeltaR between electron and generator particle matched to electron"),
             binsX = cms.untracked.vdouble(300,0,6),
             inputVariables = cms.vstring("genMatchedParticle.dR"),
-        )
+        ),
         cms.PSet (
             name = cms.string("electronGenPt"),
             title = cms.string("Gen Electron Transverse Momentum;Gen electron p_{T} [GeV]"),
--- a/Configuration/python/histogramDefinitions.py
+++ b/Configuration/python/histogramDefinitions.py
@@ -183,12 +183,6 @@ MuonHistograms = cms.PSet(
             inputVariables = cms.vstring("abs (genMatchedParticle.bestMatchPdgId)"),
             ),
         cms.PSet (
-            name = cms.string("muonBestMatchDeltaR"),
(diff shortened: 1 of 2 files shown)


jalimena:
Tested in our mumu channel on a few events. Given the histogram below (and also printouts from InfoPrinter), this PR works as expected - notice the stats box.

Screen Shot 2020-02-10 at 10 12 21 AM

jalimena:
i believe this PR can be merged

btcardwell:
Looks good to me, though it's probably best for Brian to take a quick look as well.

Configuration/python/histogramDefinitions.py
```diff
@@ -522,6 +528,12 @@
             binsX = cms.untracked.vdouble(getPdgBins(["unmatched", "quarks", "leptons", "bosons"])),
             inputVariables = cms.vstring("abs (genMatchedParticle.bestMatchPdgId)"),
             ),
+        cms.PSet (
+            name = cms.string("electronBestMatchDeltaR"),
+            title = cms.string(";#DeltaR between electron and generator particle matched to electron"),
+            binsX = cms.untracked.vdouble(300,0,6),
+            inputVariables = cms.vstring("genMatchedParticle.dR"),
+        )
```
btcardwell: Missed a comma here

jalimena: thanks, fixed now!

jalimena:
ping @bpf6qc

bpf6qc:
Sorry for the delay in responding to this. I have some comments.

1. `dR` should be initialized to INVALID_VALUE for instead of zero, since zero could indicate a valid match.
2. `dR` should be named instead `bestMatchDR` or similar to indicate that it's the dR only to the `bestMatch`.
3. A bigger question -- do we need this? `genMatchedParticle.bestMatchPdgId` seems like a convenience already given that you could do `genMatchedParticle.bestMatch.pdgId` and equivalently:

```deltaR(muon.genMatchedParticle.bestMatch, muon)```

Although as I type that out, I see this could be confusing with an inputCollection of `("muons, "muons")` and looping over things twice.

In short for this third one if the comment is "we'd like this convenience", I'm happy to merge this.

jalimena:
thanks, @bpf6qc . 

1 and 2 should now be done.

for 3, I believe we tried what you suggest, but it didn't work. I didn't see another way to do this (and believe me, I tried)

PR #152, opened 2020-02-26
Gen matching update

jalimena:
This PR:
- adds the category of matching called "noFlags". That is, no gen flags from 
https://github.com/cms-sw/cmssw/blob/master/DataFormats/PatCandidates/interface/PackedGenParticle.h#L459-L465
This is ok, however, because GenMatchable by default looks at the packedGenParticles collection (all status 1 particles). This should be the default configuration used by displaced leptons.

- renames bestMatch to something more descriptive, namely, "promptOrTauDecay"

- change the main histograms to point to the noFlags version

This PR has been tested and accurately matches events and makes histograms as expected.

### DIFF
--- a/Collections/interface/GenMatchable.h
+++ b/Collections/interface/GenMatchable.h
@@ -37,13 +37,20 @@ namespace osu
             edm::Ref<vector<osu::Mcparticle> > directHardProcessTauDecayProductFinalState;
 
             //is particle final state and (prompt or direct decay product of prompt tau)
-            edm::Ref<vector<osu::Mcparticle> > bestMatch;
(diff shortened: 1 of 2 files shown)


jalimena:
hold a sec on this PR, @btcardwell  and I are checking one thing that we could theoretically add/fix to the histograms

jalimena:
nevermind, we can't see a way to do it easily, so if you are ok with this PR the way it is, @bpf6qc , we think we can merge it now. thoughts?

bpf6qc:
Sorry @jalimena, I didn't see your comment yesterday. I am fine with these changes and will merge now.

jalimena:
no worries, thanks!

PR #153, opened 2020-02-26
Produce electron VIDs when tracks are used

bpf6qc:
@btcardwell I was slightly wrong in #149, the track producer runs a copy of the electron VID producer and needs to be considered.

This PR adds VID producers if either electrons or tracks are used.

### DIFF
--- a/Configuration/python/processingUtilities.py
+++ b/Configuration/python/processingUtilities.py
@@ -939,7 +939,7 @@ def set_input(process, input_string):
         return
 
 def customizeMINIAODElectronVID(process, collections, usedCollections):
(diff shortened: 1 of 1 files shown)


jalimena:
seems reasonable. have you checked that this one change is all that's needed to get what you need? have you run the changes and confirm you're happy?

btcardwell:
I agree that this looks good and would be happy to merge if you're ready @bpf6qc

bpf6qc:
it definitely fixes my issue, without this it complains violently about missing products and a useful value gets a nonsense value. With it, everything runs as it should.

PR #154, opened 2020-03-05
Leave only example histograms in histogramDefinitions.py

bpf6qc:
All histograms are removed, and some tutorial-styled examples are left. The PSet names are changed to avoid conflicts with other analyses' PSets.


bpf6qc:
Moving these to your analysis repo should be very simple, just copy/paste the existing contents into some new python config file. If you were importing this before, then nothing in your repo should conflict with these names and you just change the file you import from.

marthaEstefany:
I did the copy/paste thing and seems to work fine in DisplacedSUSY. I did a quick interactive test to be sure that I "imported" everything correctly. 
@jalimena @btcardwell can you skim my latest commit?

jalimena:
hi @marthaEstefany , can you clarify what your "quick interactive test" entailed? I think you should
- run over a few events, with your changes imposed, and see that you get all the histograms you expect
- compare that with histograms run before your changes, to see that all the histograms you want are there and filled appropriately

I think that should be sufficient to prove we are ok.

marthaEstefany:
hi @jalimena . My interactive test was, for the most part, as you suggested. 

- I ran over a few events interactively, in the EEchannel. Particularly, I ran events through our preselection criteria, defined for this channel (EE).
-  The output, hist.root, from cmsRun looked normal. The plots are filled as I would expect - nothing looks strange. 

Comparing this output to what I would have gotten before the change is a bit hard at this point (i'm not sure how to go backwards in my area) but I can run over a skim and make a quick comparison. 

thanks.

bpf6qc:
@marthaEstefany one thing you could look at is this:
https://godoc.org/go-hep.org/x/hep/groot/cmd/root-diff

You're just moving the same histograms from one file to another, so success here means
1) it still runs, meaning everything is still defined and imported properly
2) it produces the exact same list of histograms as before, meaning nothing changed

marthaEstefany:
@bpf6qc thanks for the link.
I'll try this way of comparing root files, to be doubly sure that it's fine and we didn't lose any histograms with the move.

marthaEstefany:
all looks good to me. 
i compared histograms produced after the move with histograms produced before the move.
All histograms seem to be accounted for, for the EE channel at least, and also filled in reasonably. 
@bpf6qc, can you do a quick skim of my changes (link below)? 
https://github.com/DisplacedSUSY/DisplacedSUSY/commit/5208c7a2fefe6443798cbc85920bfb2acc06baf5

bpf6qc:
@marthaEstefany this looks correct to me, no instance of the framework file is left in your code so if you have forgotten something, you will find an error to remind you. I'll leave it to someone to approve/merge this PR when they're ready (@jalimena @btcardwell as well)

PR #155, opened 2020-03-12
Add muon rho

btcardwell:
Add rho variable (fixedGridRhoFastjetAll) to muon object for new muon isolation definition in Displaced Leptons. All other needed changes can be implemented in our analysis repo. I have tested this PR and believe it to be ready to merge.

### DIFF
--- a/Collections/interface/Muon.h
+++ b/Collections/interface/Muon.h
@@ -19,6 +19,7 @@ namespace osu
         Muon (const TYPE(muons) &, const edm::Handle<vector<osu::Mcparticle> > &, const edm::ParameterSet &, const osu::Met &);
         ~Muon ();
 
+        const float rho() const;
         const double pfdBetaIsoCorr () const;
         const double sumChargedHadronPtCorr () const;
         const double sumPUPtCorr () const;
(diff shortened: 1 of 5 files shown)


jalimena:
hi @btcardwell , is "fixedGridRhoFastjetAll" the name for it in all Run2 configurations? all years/ releases? 

also, how have you tested this PR? would it work if say the input file was AOD?

btcardwell:
hi @jalimena, yes, fixedGridRhoFastJetAll is the name for all relevant years/releases. I have only tested this PR with miniAOD input, but the implementation is exactly analogous to the implementation for electrons (and jets and photons), so any issues from this PR would be seen in the other objects as well.

jalimena:
ok thanks, @btcardwell !

@bpf6qc : I think this PR is good to go, but maybe you should have a quick look first. If you're happy, then please merge it.

bpf6qc:
@btcardwell @jalimena `fixedGridRhoFastJetAll` is used in other framework object producers so this has the same dependency on name. There is protection against failures with `if(rho.isValid())` which I see was added here.

If the collection name or recommendation changes in the future, this will need to be changed, but I am very okay with how it's used here.

Merging now.

PR #156, opened 2020-03-26
Add uniqueMotherPdgId to Mcparticle

btcardwell:
Add uniqueMotherPdgId to Mcparticle following the implementation in HardInteractionMcparticle. This works as expected when tested in 10_2_12.

### DIFF
--- a/Collections/interface/Mcparticle.h
+++ b/Collections/interface/Mcparticle.h
@@ -13,7 +13,15 @@ namespace osu
         Mcparticle ();
         Mcparticle (const TYPE(mcparticles) &);
         ~Mcparticle ();
+
+	const int uniqueMotherPdgId () const;
+
+	void set_uniqueMotherPdgId (double value) { uniqueMotherPdgId_  = value; };
+
+	int uniqueMotherPdgId_;
+
     };
+
 }
 
 #else
--- a/Collections/plugins/HardInteractionMcparticleProducer.cc
+++ b/Collections/plugins/HardInteractionMcparticleProducer.cc
@@ -30,8 +30,8 @@ HardInteractionMcparticleProducer::produce (edm::Event &event, const edm::EventS
     pl_->emplace_back (object);
     osu::HardInteractionMcparticle &hiMcpart = pl_->back();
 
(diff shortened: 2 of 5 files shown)


jalimena:
also, it would be good if @bpf6qc  can also take a look and give his 2 cents

btcardwell:
Hey, we actually don't need to add anything to genMatchable because the object we get back when using genMatchable is an Mcparticle. It looks like `genMatchedParticle.noFlags.uniqueMotherPdgId` gives us exactly what we want.

jalimena:
excellent. @bpf6qc , ok to merge then?

bpf6qc:
This is okay to merge for me, and yes genMatchable provides a reference to this object so no more modifications are needed.

From a style perspective:
```
if(!uniqueMother(mcpart)) continue;
mcpart.set_uniqueMotherPdgId(uniqueMother(mcpart)->pdgId());
```
is a bit inefficient for two reasons. Firstly calling `uniqueMother` twice is looping again, and secondly the `continue` could be a trap if someone ever adds code below this in the same brackets. It's not _that_ bad, I'm torn on mentioning it or not. I'll leave it to others' opinion.

Otherwise go ahead and merge this.

btcardwell:
Not a bad point on the style issue. I haven't tested it, but would something like the following address both your concerns?

`const reco::Candidate *uniqueMother = uniqueMother(mcpart);`
`if (uniqueMother(mcpart)) mcpart.set_uniqueMotherPdgId(uniqueMother->pdgId());`

bpf6qc:
@btcardwell yes exactly. There might be the same code in HardInteractionMcparticle too, maybe it's a good chance to change this.

jalimena:
ok cool, agreed. I'd ask @btcardwell to make this change for both mcparticles and hardInteractionmcparticles.

btcardwell:
Alright, I changed mcparticle and hardinteractionmcparticle and tested again. If everyone's on board, I believe this should be good to go.

bpf6qc:
Looks good by me, thanks for the change.

jalimena:
+1

PR #161, opened 2020-06-04
add EventSetup to EventVariableProducers

jalimena:
Added the edm::EventSetup to AddVariables in the EventVariableProducer

For example, useful for getting the L1 information in an event variable producer with 
```l1GtUtils_->retrieveL1(event,setup,algToken_);```

This PR will require everyone to make similar changes for all of their event variable producers in their respective packages.

Tested locally with the DisplacedSUSY event variable producer: can indeed retrieve the L1 info.

### DIFF
--- a/AnaTools/interface/EventVariableProducer.h
+++ b/AnaTools/interface/EventVariableProducer.h
@@ -32,7 +32,7 @@ class EventVariableProducer : public edm::EDFilter
 
       // Methods
 
-      virtual void AddVariables(const edm::Event &) = 0;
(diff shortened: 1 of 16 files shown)


jalimena:
hey, i think this should be a fairly simple PR. could anyone look at it soon? i would like to have this merged before i push related commits to DisplacedSUSY and run the jobs. but again, tested locally and it works fine

btcardwell:
This all looks fine to me, but I think @bpf6qc is more likely to have have thoughts on implementation. Is this all good from your perspective Brian?

btcardwell:
Great! I'll go ahead and merge then.

PR #162, opened 2020-06-19
Add beamspot and gen-level lepton track reference point information to electrons and muons

btcardwell:
This has been tested in CMSSW 10_2_12 and works as expected. I believe it can be merged as is, though if it makes sense to others, I can first add the minor improvement described below.

I notice that a small fraction of the time, the matched gen particle is really low pT. I haven't looked into it yet, but does anyone know if the prunedGenParticles is pT ordered? If so, it might be smarter to iterate backwards through the collection so the highest-pT match is saved instead of the lowest.

### DIFF
--- a/Collections/interface/Electron.h
+++ b/Collections/interface/Electron.h
@@ -24,6 +24,10 @@ namespace osu
         const double sumChargedHadronPtCorr () const;
         const double sumPUPtCorr () const;
         const int electronPVIndex () const;
+        const double genVx () const;
+        const double genVy () const;
+        const double genPx () const;
+        const double genPy () const;
         const double genD0 () const;
         const double d0SmearingVal () const;
         const float dEtaInSeed () const;
@@ -53,6 +57,10 @@ namespace osu
         void set_sumChargedHadronPtCorr (double value) { sumChargedHadronPtCorr_  = value; };
         void set_sumPUPtCorr (double value) { sumPUPtCorr_  = value; };
         void set_electronPVIndex (int value) { electronPVIndex_  = value; };
+        void set_genVx (double value) { genVx_  = value; };
+        void set_genVy (double value) { genVy_  = value; };
+        void set_genPx (double value) { genPx_  = value; };
+        void set_genPy (double value) { genPy_  = value; };
         void set_genD0 (double value) { genD0_  = value; };
         void set_d0SmearingVal (double value) { d0SmearingVal_  = value; };
         void set_passesTightID_noIsolation_LegacySpring15 (const reco::BeamSpot &, const TYPE(primaryvertexs) &, const edm::Handle<vector<reco::Conversion> > &);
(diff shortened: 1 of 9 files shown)


jalimena:
This PR looks fine to me.

Indeed, I believe all collections should be pt ordered, but I'm pretty sure that the order goes highest to lowest, not the other way around. You can of course print out the pts to check, but I think the code is already doing what you suggest.

jalimena:
Bryan and I chatted: the loop is likely to go highest to lowest pt, but then if there is more than 1 possible candidate match, the lowest pt one would get saved. Bryan is going to implement a break instead (and not change the order, of course). Looks like a break was missing here by mistake.

bpf6qc:
I'll be annoying here and come from an overall design perspective, so override this if the group desires. I don't feel we should be specializing object producers in this way, where the framework should more generally provide a mechanism. Specifically what you seem to need is an analogue of `GenMatchable` that works with `hardInteractionMcparticles`/`prunedGenParticles`.

Some issues/comments:

1. dR < 0.1 is hard-coded, as is the use of prunedGenParticles
2. Only px/py are stored for example, but providing access to the vertex and p4 would enable much more analysis
3. For reconstructed electrons that are not matched to gen-electrons, no information is stored; a general user would expect "electron.gen_d0" to be as named in any case
4. All of this can be replicated in python if you had the gen-matched object

I can still appreciate "we just need to finish our analysis" here as this PR enables something, so I can see merging this. But my suggestion is a `GenMatchable` analogue.

Even farther afield, much more work to be done later, I recognize adding more and more to object constructors like this gets the framework very bloated (and it already is). Perhaps for gen-matching, we could instead produce an `Association` collection between objects and packed/pruned matches, and only produce the association map when `genMatched` is used in the channel.

jalimena:
Hi @bpf6qc , all of this makes sense in general. I think Bryan can fairly easily implement your #1-3 now. I think the rest should be done down the road, once we finish this analysis.

btcardwell:
yeah, I'm definitely with you on a genMatchable analogue being better, @bpf6qc, but I think I'm leaning towards "we just need to finish our analysis" in this case. I could partly address issue #1 by adding a default dR value in collectionProducer_cff.py, but I don't think I see simple ways to address the other shortcomings. What do you think?

bpf6qc:
@btcardwell The issues/comments I listed were intertwined with my overall comment. As a matter of enabling your analysis, these aren't issues for me at all.

For a simple suggestion now, how about you add a python parameter like `genD0DR` (or whatever name) to replace `0.1`, and secondly add a comment saying this is a temporary solution that probably can be solved by extending GenMatchable to pruned?

btcardwell:
Sounds great. Will do.

jalimena:
the comment in the code could also refer back to this PR, with the full list of improvements we want to make. or better yet, we make this into a github issue

btcardwell:
ok, I believe this is ready to merge. I'll make an issue as well, as @jalimena suggested.

bpf6qc:
@btcardwell This looks fine to me then. I'll approve and leave it to @jalimena to merge, since I haven't done any testing.

PR #164, opened 2020-06-22
add d0 wrt beamspot calculations for tests with vertex=(0,0,0)

jalimena:
this is for tests for displaced leptons. this will give you the wrong gen d0 calculation, because here, the vertex is set to (0,0,0). but this is useful for a quick test for us.


jalimena:
tested and working fine in 10_2_12

bpf6qc:
This won't affect anything, and is of course mergeable on that merit. But I don't see this as necessary in the framework repo -- it can be defined in DisplacedSUSY and imported anywhere. `muonD0WRTBeamspot` is for example relevant for many analyses, is there a reason to add this testing string here?

jalimena:
hi @bpf6qc , ok sure, i just didnt have a particularly good place to put it in DisplacedSUSY. but i can make one.

PR #169, opened 2020-07-23
Add automatic sanity check for cut and histogram definitions

btcardwell:
Add check to add_channels that raises an exception if any cuts or histograms have invalid attributes. This protects against typos like `IndexX` or `numberrequired` that currently pass silently.

### DIFF
--- a/Configuration/python/processingUtilities.py
+++ b/Configuration/python/processingUtilities.py
@@ -255,10 +255,6 @@ def add_channels (process,
         'alias',
         'numberRequired',
         'isVeto',
-        '_isFrozen',
-        '_ParameterTypeBase__isTracked',
-        '_isModified',
-        '_Parameterizable__parameterNames',
     ]
     validHistAttributes = [
         'name',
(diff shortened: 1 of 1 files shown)


jalimena:
+1
highly support. i can't think now of any other checks we could make. i leave it to @bpf6qc  to merge, if he's happy.

btcardwell:
I should add that this has been tested and works as expected in 10_2_X for all DisplacedSUSY cuts and hists.

bpf6qc:
I think this is a good idea, but I balk at the PSet defaults and built-ins, the ones beginning with `_`, since there may be CMSSW updates or Python built-ins that show up. I'd like to see only the 'regular' attributes checked, and anything beginning with `_` ignored in this.

btcardwell:
Good point @bpf6qc. I've pushed another commit to ignore attributes starting with `_`.

bpf6qc:
Thanks for the change @btcardwell, this looks good to merge for me.

PR #170, opened 2020-08-06
add MET_2017_withoutB dataset

jalimena:
- needed to find trigger efficiency for displaced leptons with an unbiased dataset. no need for 2017B, when some of the triggers in the analysis didnt yet exist
- should be a pretty straightforward PR
- tested in CMSSW_9_4_8


bpf6qc:
Since this is a composite dataset, it should be defined in all the same places as `'MET_2017'`. You'll need to add it to `datasets` and remove it from `maxEvents`.

jalimena:
ok, done

PR #171, opened 2020-08-06
Json file check

jalimena:
I recently made a mistake when submitting jobs, such that I used the wrong JSON file - I accidentally used a 2018 JSON file when running over 2017 data in 9_4_8. This kind of mistake can be easily avoided with the check I've put in, in this PR, based on what CMSSW release you're in. I made it as a check, and the user can override it, because I can imagine that in the future, we won't have this strict categorization on data from a given year corresponding directly to only one CMSSW release, and also because we do some weird things with this repo and maybe this isn't universally applicable.

Sorry for the other OSUT3Analysis commit that doesn't relate - as you'll see, I reverted it.

I've tested this PR, and it works as I expect.


bpf6qc:
This is fine but it misses the case where no 16/17/18 is provided and 2015 is used, or if some disturbed person ever calls like "1718". Could you change this block to a more `elif` based logic, and also check 2015 for 7_6_X?

Also in the long run, the direction of JSON/year logic works for this. CMSSW_1x_ could be used for 2+ years for example, but those 2 years' JSONs will each have a single appropriate release.

jalimena:
thanks, @bpf6qc . does my new commit satisfy your request?

bpf6qc:
I was a bit vague, one more change unfortunately. 2015 has been like `R_Silver` without any numbers, so:
`if re.search('15$', arguments.JSONType):`
that block needs to be at the end with just `else:` (and then the 2016 at the top with just `if`)

jalimena:
ah right, thanks for explaining. i think should be good now

jalimena:
I think this PR is fine for the moment, could definitely be improved in the future. But better than what we had before. My tests show I haven't made a mistake, so merging now. thanks, @bpf6qc

PR #173, opened 2020-08-12
Make efficiency plot with ratio histogram, without resorting to TH1

jalimena:
This PR:

- Makes the efficiency plot with a ratio histogram, without resorting to plotting a TH1 for the efficiency curve. This allows you to have the correct error bars on the efficiency plot, while still having the ratio plot.
- As a result of the above, the --noTGraph option is no longer needed, so this PR removes it.
- This PR also switch to TEfficiency from TGraphAsymmErrors, since that is supposed to be a better class to use. Has no effect on the resulting plot though.
- Improve cosmetics of the plot

Tested in 9_4_8, produces these example output plots:

[muonPt.pdf](https://github.com/OSU-CMS/OSUT3Analysis/files/5062675/muonPt.pdf) - an example with the ratio plot
[muonLeadingPt.pdf](https://github.com/OSU-CMS/OSUT3Analysis/files/5062676/muonLeadingPt.pdf) - an example without the ratio plot

### DIFF
--- a/Configuration/scripts/makeEfficiencyPlots.py
+++ b/Configuration/scripts/makeEfficiencyPlots.py
@@ -32,8 +32,6 @@ parser.add_option("-E", "--ratioRelErrMax", dest="ratioRelErrMax",
                   help="maximum error used in rebinning the ratio histogram")
 parser.add_option("-A", "--addOneToRatio", action='store_true', dest="addOneToRatio",
                   help="add one to the ratio so that data/MC is plotted")
-parser.add_option("--noTGraph", action="store_true", dest="noTGraph", default=False,
-                  help="don't make a TGraph; just make a TH1")
 parser.add_option("--ylog", action="store_true", dest="setLogY", default=False,
                   help="Set logarithmic scale on vertical axis on all plots")
 parser.add_option("--ymin", dest="setYMin",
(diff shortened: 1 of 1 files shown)


jalimena:
gentle ping

Configuration/scripts/makeEfficiencyPlots.py
```diff
@@ -262,11 +266,24 @@ def MakeOneHist(dirName, histogramName):
         if arguments.normalizeToUnitArea:
             yAxisLabel = yAxisLabel + " (Unit Area Norm.)"
 
-        #Histogram = ROOT.TGraphAsymmErrors(NumHistogramObj,DenHistogramObj)
-        if arguments.noTGraph or Histogram.Class().InheritsFrom("TH2"):
+        #check if bin content is consistent
+        for i in range(1,Histogram.GetNbinsX()+1):
```
bpf6qc, line 270: I think this should be more verbose if failed, the user should know their efficiency is odd. You might use `TEfficiency::CheckConsistency` or `CheckBinning` which come with info printouts.

Configuration/scripts/makeEfficiencyPlots.py
```diff
@@ -375,6 +388,10 @@ def MakeOneHist(dirName, histogramName):
     if arguments.plot_hist:
         plotting_options = "HIST"
 
+    h = TH2F("h1","",NBins[0],MinXValues[0],MaxXValues[0],110,0.,1.1)
```
bpf6qc, line 391: There's an interaction here with `finalMax` ~50 lines above, where the y-max will be 1.1 even if `setYMax` is called.

On line 340: `finalMax = 1.1` (then 341-342 can override)
On line 391 `1.1` --> `finalMax`

Configuration/scripts/makeEfficiencyPlots.py
```diff
@@ -262,11 +266,24 @@ def MakeOneHist(dirName, histogramName):
...
+        Nbins = HistogramClone.GetNbinsX()
+        MaxXValue = HistogramClone.GetXaxis().GetBinLowEdge(Nbins+1)
+        MinXValue = HistogramClone.GetXaxis().GetBinLowEdge(1)
+
+        #this Histogram becomes the main TEfficiency
+        if Histogram.Class().InheritsFrom("TH2"):
             Histogram.Divide(DenHistogram)
         else:
-            Histogram = TGraphAsymmErrors(Histogram,DenHistogram)
+            Histogram = TEfficiency(Histogram,DenHistogram)
```
bpf6qc, line 286: I am not recommending a code change here, but in case someone has specialized needs we should remember that these divide functions use default methods:

https://root.cern.ch/doc/master/classTEfficiency.html (c.f. section IV)

Perhaps just a link to that in comment is enough.

Configuration/scripts/makeEfficiencyPlots.py
jalimena: done

Configuration/scripts/makeEfficiencyPlots.py
jalimena: done, good catch

Configuration/scripts/makeEfficiencyPlots.py
jalimena: added this comment as you suggest. as i wrote in the comment, the default method should usually be the right one, at least as far as the uncertainties are concerned.

jalimena:
thanks!

PR #175, opened 2020-08-28
make framework work with AOD

jalimena:
This PR allows the framework to work with AOD (and AODSIM) data formats. The collectionMapAOD and DataFormatAOD.h needed to have a few additions, a few more files needed to be included, and some parts of the code only work if the format is not AOD or if trigobjs is valid.

**Tests I did:** 
Compiles in 8_0, 9_4, and 10_2, 

with both the 
`AnaTools/scripts/setupFramework.py -f AOD`
and the 
`AnaTools/scripts/setupFramework.py -f MINI_AOD` or `MINI_AOD_2017`
options.

Can run over an AOD file in 8_0 and make histograms and skims.

**There are three warnings when I compile with -f AOD that I'm not sure how to fix, help welcome:**

1.
`>> Compiling  /uscms_data/d3/alimena/DisplacedLeptons/CosmicMC/CMSSW_10_2_12/src/OSUT3Analysis/AnaTools/src/EventVariableProducer.cc 
In member function 'anatools::TypeWithDict anatools::TypeWithDict::toType() const':
cc1plus: warning: 'void* __builtin_memset(void*, int, long unsigned int)': specified size 18446744073709551608 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]`

2.
`/uscms_data/d3/alimena/DisplacedLeptons/CosmicMC/CMSSW_9_4_8/src/OSUT3Analysis/Collections/plugins/OSUTauProducer.cc: In member function 'virtual void OSUTauProducer::produce(edm::Event&, const edm::EventSetup&)':
/uscms_data/d3/alimena/DisplacedLeptons/CosmicMC/CMSSW_9_4_8/src/OSUT3Analysis/Collections/plugins/OSUTauProducer.cc:49:17: warning: unused variable 'tau' [-Wunused-variable]
       osu::Tau &tau = pl_->back ();
                 ^~~
`

3.
`/uscms_data/d3/alimena/DisplacedLeptons/CosmicMC/CMSSW_9_4_8/src/OSUT3Analysis/Collections/plugins/OSUGenericJetProducer.cc: In instantiation of 'void OSUGenericJetProducer<T>::produce(edm::Event&, const edm::EventSetup&) [with T = osu::Jet]':
/uscms_data/d3/alimena/DisplacedLeptons/CosmicMC/CMSSW_9_4_8/src/OSUT3Analysis/Collections/plugins/OSUGenericJetProducer.cc:378:34:   required from here
/uscms_data/d3/alimena/DisplacedLeptons/CosmicMC/CMSSW_9_4_8/src/OSUT3Analysis/Collections/plugins/OSUGenericJetProducer.cc:161:10: warning: unused variable 'jet' [-Wunused-variable]
       T &jet = pl_->back ();
          ^~~
`

### DIFF
--- a/AnaTools/interface/CommonUtils.h
+++ b/AnaTools/interface/CommonUtils.h
@@ -226,12 +226,14 @@ namespace anatools
 
   template<class T> bool jetPassesTightLepVeto (const T &);
 
+#if IS_VALID(trigobjs)
   template<class T> bool isMatchedToTriggerObject (const edm::Event &, const edm::TriggerResults &, const T &, const vector<pat::TriggerObjectStandAlone> &, const string &, const string &, const double = 0.1);
   bool getTriggerObjects (const edm::Event &, const edm::TriggerResults &, const vector<pat::TriggerObjectStandAlone> &, const string &, const string &, vector<const pat::TriggerObjectStandAlone *> &);
   bool getTriggerObjectsByFilterSubstring (const edm::Event &, const edm::TriggerResults &, const vector<pat::TriggerObjectStandAlone> &, const string &, const string &, vector<const pat::TriggerObjectStandAlone *> &, const vector<string> & = vector<string> ());
   template<class T> const pat::TriggerObjectStandAlone *getMatchedTriggerObject (const edm::Event &, const edm::TriggerResults &, const T &, const vector<pat::TriggerObjectStandAlone> &, const string &, const string &, const double = 0.1);
   bool triggerObjectExists (const edm::Event &, const edm::TriggerResults &, const vector<pat::TriggerObjectStandAlone> &, const string &, const string &);
   bool passesL1ETM (const edm::Event &, const edm::TriggerResults &, const vector<pat::TriggerObjectStandAlone> &, double &);
+#endif
 
   void logSpace (const double, const double, const unsigned, vector<double> &);
   void linSpace (const double, const double, const unsigned, vector<double> &);
@@ -283,6 +285,7 @@ anatools::jetPassesTightLepVeto (const T &jet)
 #endif
 }
 
+#if IS_VALID(trigobjs)
 template<class T> bool
 anatools::isMatchedToTriggerObject (const edm::Event &event, const edm::TriggerResults &triggers, const T &obj, const vector<pat::TriggerObjectStandAlone> &trigObjs, const string &collection, const string &filter, const double dR)
 {
@@ -361,4 +364,5 @@ anatools::getMatchedTriggerObject (const edm::Event &event, const edm::TriggerRe
   return NULL;
 }
 
+#endif
 #endif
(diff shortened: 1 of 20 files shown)


jalimena:
gentle ping

bpf6qc (commented):
In several places, there are additions like `#if DATA_FORMAT == AOD` or similar specific comparisons. It would be better to use things like `#if DATA_FORMAT_FROM_AOD` or `#if DATA_FORMAT_FROM_MINIAOD`, for reasons of readability (it says clearly to future editors that this is MINIAOD code and is used in that data) and support of future changes (there's already MINIAOD_2017 with small compilation changes)

Secondly, are we sure we want to disable all the `trigobjs` utility functions? I was going to suggest instead of blocking them out with `#if IS_VALID(trigobjs)`, changing out the argument types:
`pat::TriggerObjectStandAlone` --> `TYPE(trigobjs)`
and then using `DATA_FORMAT_FROM_MINIAOD` etc to switch behavior. But looking at `PATTriggerProducer`, it looks directly at `gtDigis`?! So, if there is no AOD cousin to `pat::TriggerObjectStandAlone`, I can support blocking these functions out entirely.

jalimena:
Thanks, @bpf6qc .

For your first comment, does my latest commit fix it appropriately?

For your second comment, yeah, I couldn't find an AOD version of pat::TriggerObjectStandAlone. Here is the event content for the AOD file I was using for testing [aodEventContent.txt](https://github.com/OSU-CMS/OSUT3Analysis/files/5175572/aodEventContent.txt) - let me know what you think we should do.

Any ideas on the warnings I don't know how to fix? See my first comment.

jalimena:
ping

bpf6qc:
Hi @jalimena, getting back to this. Your changes do address my comment, thanks. For compiler warning 2 and 3, it looks like the object references are used only in certain cases. You can fix these by just moving the reference line into the pre-compiler macro check. So

OSUTauProducer.cc line 49 `osu::Tau &tau = pl_->back ();` -- just move this into the `#if IS_VALID(trigobjs)` block below it.

OSUGenericJetProducer.cc line 161 `T &jet = pl_->back ();` -- this is slightly more obscure. If `STOPPED_PTLS` is _not_ defined, and also the data format is _not_ from MINIAOD, then it's never used. You could change this to be...

```# starting at line 157
for (const auto &object : *collection)
    {
#ifndef STOPPPED_PTLS
      pl_->emplace_back (object, particles, cfg_);
#else // STOPPPED_PTLS
      pl_->emplace_back (object);
#endif

#if DATA_FORMAT_FROM_MINIAOD

      T &jet = pl_->back ();

      // medianLog10(ipsig) CALC
      std::vector<double> ipsigVector;
```

If you could please make these changes and test them quickly, that would be great! Compiler warning 1 is too obscure for me right now, and appeared when the GCC version went up. I'll have to look into that in another pull.

jalimena:
thanks @bpf6qc ! done. from my side, I believe this is ready to be merged.

bpf6qc:
Thanks for the additional changes @jalimena, I'll merge now.

bpf6qc (approved):

PR #177, opened 2020-09-14
Fix ratio plot y-axis range bug in makePlots

btcardwell:
Fix a bug that lead to incorrect ratio plot y-axis ranges. Also fixed a couple related value comparisons with `is`. This works in CMSSW_10_2_12.

### DIFF
--- a/Configuration/scripts/makeComparisonPlots.py
+++ b/Configuration/scripts/makeComparisonPlots.py
@@ -515,7 +515,7 @@ def MakeOneDHist(histogramDirectory, histogramName,integrateDir):
                 makeRatio = functools.partial (ratioHistogram,Histogram, Reference)
                 if arguments.ratioRelErrMax is not -1: # it gets initialized to this dummy value of -1
                     makeRatio =  functools.partial (makeRatio, relErrMax = float(arguments.ratioRelErrMax))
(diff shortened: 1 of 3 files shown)


jalimena (approved):
ok this looks good for now. i'm merging it now because i think it's not controversial and we would like to use it now for the analysis. @bpf6qc , if you have problems with this PR, we can make a new one later to address your concerns

bpf6qc:
So this works, but it doesn't fix the underlying weird python that was the underlying problem -- c.f. lines 425-427 in `makeComparisonPlots.py`:
```
addOneToRatio = -1
    if arguments.addOneToRatio:
        addOneToRatio = arguments.addOneToRatio
```
In this case arguments.addOneToRatio is controlled by `action='store_true'` with `-A`, so it should never be false. It's just weird python to convert to `-1` and not supply a default for `-A`.

btcardwell:
Yeah, I'm with you on this being weird bad python. I was hoping to minimize my changes and avoid falling into the task of fixing a bunch of weird bad python in these early osut3analysis scripts. Do you think we should open a new PR to make this less weird?

bpf6qc:
@btcardwell on this, I'm split -- maybe I cared enough to point it out, but changes tend to cause bugs. I'll leave it to you to decide.

PR #178, opened 2020-10-14
Update DisplacedSUSY lifetime reweighting rule to use privately produced 0.1 mm samples

btcardwell:
Simple change to use privately produced 0.1mm samples instead of reweighting 1mm samples.  The current implementation will use the 1mm sample as the source for ctaus from 0.2-1mm and the 0.1mm sample for everything below that. @jalimena, we should make sure we agree on this choice before merging.

### DIFF
--- a/Configuration/python/configurationOptions.py
+++ b/Configuration/python/configurationOptions.py
@@ -9104,7 +9104,7 @@ for index, sample in enumerate(signal_datasets):
 
     sourceCTau = 0.1 * 10**(math.ceil(math.log10(float(lifetime(sample)))))
     # special case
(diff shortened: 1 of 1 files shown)


jalimena:
@btcardwell and i chatted, and we're now on the same page. the way he has it, we think, is right and best. so i'll approve this PR, and wait just a few days for @bpf6qc to comment before merging

jalimena (approved):

bpf6qc:
I don't see why this needs to be expressed as a product, but a trivial comment won't hold up merging! Especially seeing the agreement for an analysis choice this is important.

PR #179, opened 2020-11-17
changes to over/underflow

jalimena:
with option to turn it off, as in makePlots.py

### DIFF
--- a/Configuration/scripts/makeComparisonPlots.py
+++ b/Configuration/scripts/makeComparisonPlots.py
@@ -46,6 +46,8 @@ parser.add_option("-g", "--generic", action="store_true", dest="generic", defaul
                   help="generic root file directory structure; does not assume that channel dir is in OSUAnalysis dir")
 parser.add_option("-s", "--signif", action="store_true", dest="makeSignificancePlots", default=False,
                                     help="Make significance plots")
+parser.add_option("--NO", "--noOverUnderFlow", action="store_true", dest="noOverUnderFlow", default=False,
+                  help="Do not add the overflow and underflow entries to the last and first bins")
 parser.add_option("-v", "--verbose", action="store_true", dest="verbose", default=False,
                   help="verbose output")
 
@@ -57,6 +59,7 @@ if arguments.localConfig:
     sys.path.append(os.getcwd())
     exec("from " + re.sub (r".py$", r"", arguments.localConfig) + " import *")
 
+noOverUnderFlow = arguments.noOverUnderFlow
 
 
 #### deal with conflicting arguments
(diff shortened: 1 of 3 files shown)


btcardwell:
This looks like a good idea and the plots look good. Are `lowLimit` and `uppLimit` used anywhere?

jalimena:
Thanks, @btcardwell ! re low/uppLimit, I had missed a part in the makePlots.py code. Now fixed. (This matters for the integral of histograms, so it wouldn't have shown up in my test plots above.)

jalimena:
Come to think of it, I think we want to do the same for makeEfficiencyPlots (before we divide the histograms), so my latest commit does that.

bpfrancis (approved):
The changes are uncontroversial, this looks good. There are two questions I want to raise, but neither are problems for me and are a matter of the group's tastes. I'll approve this PR and pose them to see how people feel:

1. I'm not sure what the default should be, to show the over/underflows or not. Generally I don't show them, but as long as users know what's happening they can see the flag to turn this on/off.

2. The on/off flag controls both the visual display of over/underflow, but also the numerical treatment of them in integral plots. I would think they should always be included in integral plots (doing the right thing with left/right directions), and be a choice to show or not in the non-integrated histograms.

bpfrancis:
Actually a third:

3. This groups overflow with underflow. Most plots will never see underflow, like your PT examples, but perhaps it's worth separating flags in case someone's surprised. I could see disTrks using -999 for "no selected jet" and wanting to turn off the spike at zero.

jalimena:
Thanks, @bpfrancis .

Regarding your points:
1. I think we should be consistent with what is done in makePlots.py, which by default plots them.
2. Ok sounds good, I will change this here (and also in makePlots.py if it's the same there, haven't yet looked).
3. This is a good point and one we should also do for makePlots.

So I'll make the changes regarding 2 and 3 above asap. Please wait for those before merging.

jalimena:
New PR summary:
- change makeComparisonPlots to plot over/underflow by default, and also for makeEfficiencyPlots (before dividing histograms)
- change to always include under&overflow in integral histograms, in makeComparisonPlots and makePlots
- split option to turn off under&overflow plotting (noOverUnderFlow option) into two options: noOverFlow and noUnderFlow, for makePlots, makeComparisonPlots, and makeEfficiencyPlots

jalimena:
ah sorry, i meant to be clearer: it would be great if you two could have another look now

bpfrancis:
My three comments have been addressed very nicely, thanks @jalimena . I still approve, perhaps @btcardwell can look and merge.

btcardwell:
Yes, everything looks good to me. I'll go ahead and merge.

PR #181, opened 2020-12-14
add overflow and underflow by default to 2D plots

jalimena:
a 2D plot before the PR:
beforePR

and after:
afterPR

Notice the top left, top right, and bottom right bins (there's no underflow in this particular plot)

### DIFF
--- a/Configuration/scripts/makePlots.py
+++ b/Configuration/scripts/makePlots.py
@@ -198,6 +198,17 @@ dir_y_bottom = 0.0
 dir_x_right  = ts_x_right
 dir_y_top    = ts_y_bottom
 
+noOverFlow = arguments.noOverFlow
+if arguments.paperConfig:
+    if 'noOverFlow' in paperHistogram:
+        noOverFlow = paperHistogram['noOverFlow']
+###############################################
+noUnderFlow = arguments.noUnderFlow
+if arguments.paperConfig:
+    if 'noUnderFlow' in paperHistogram:
+        noUnderFlow = paperHistogram['noUnderFlow']
+###############################################
+
 
 ##########################################################################################################################################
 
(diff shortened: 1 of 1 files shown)


btcardwell:
Hmm, I'm not sure what I think of this. Do other people handle 2D overflow in this same way? I can't think of a better approach though (other than including the stats box).

jalimena:
It's not really conceptually different from what we're doing in the 1D case though, right? Matthew explicitly asked us to do this for these plots, and then I thought it should be consistent with what we're doing elsewhere in this script.

bpfrancis:
Maybe @btcardwell is thinking the same as I am -- conceptually the behavior is the same for 1D/2D, by default overflow is shown in the edge bins and a flag is provided to override this. As a matter of figure presentation, personally, I would expect a default 1D plot includes over/underflow yet a 2D plot wouldn't.

I think however, consistency in the code is more important. So I approve of this change. Maybe as a middle ground, a single (not for every plot!) print statement could be added like

`Adding over/underflow to last bins...`

And users will understand this.

jalimena:
ok, added such print statements

bpfrancis (approved):
Reminding users this is the case, these changes are good. Perhaps @btcardwell can comment and/or merge, but I approve.

btcardwell:
All sounds good. I'm still a little weirded out by the presentation*, but I agree with the goals of consistency between 1D and 2D plots and directly addressing Matthew's reqeust. I will merge.

*I think it weirds me out because in the 1D case, it's easy to unambiguously communicate the the meaning of the overflow bin (e.g. "the last bin contains the number of events with value > the last bin lower bound"), but the 2D case is trickier (e.g. "this bin contains the number of events with values x and y in the bin range but also the events with x value > the x-axis upper bound and y value < the y-axis upper bound" for the lower-right overflow bin).

PR #182, opened 2020-12-31
add H to SS to 4L samples for displaced leptons

jalimena:


jalimena:
I've run over the samples in all three releases to get the PU reweighting histograms, and everything looks fine. I've also checked that the lifetime reweighting works as I expect - no changes are needed in LifetimeReweightingProducer for these samples.

btcardwell:
This all seems good and noncontroversial to me. I'll apporove and maybe @bpfrancis can merge after he takes a look.

btcardwell (approved):

jalimena:
ping @bpfrancis

bpfrancis (commented):
Perhaps I'm missing something, but the prevalence of string.find() is not good here. Especially when `signal_datasetsH` is defined as starting with `HToSSTo4L` and is searched many times for this prefix.  `massH(sample` could be defined as just `return sample[9:sample.find('_')]`.

An alternative could be:

```
signal_datasetsH = { (mH, mS, ctau) : 'HToSSTo4L%s_%s_%smm' % (mH,mS,ctau) for mH in massesH for mS in massesS for ctau in lifetimesS }

datasets.extend( list(signal_datasetsH.values()) )
...
for ((mH, mS, ctau), sample) in signal_datasetsH.items():
    labels[sample] = 'XYZ' % (mH, mS, ctau)
    crossSections[sample] = signal_crossSectionsH[mH]
```

etc. But please find a way to do this without using `find` as often.

jalimena:
Hi @bpfrancis , 

I'm trying to do what you ask, but 
1. I'm not sure how to do this avoiding find (except for the example you give)
2. If I change signal_datasetsH as you suggest, and then print out the dataset names to check it's doing the right thing, it's really hard to follow and be sure I've got the right things.

I know we want the code to be as good as possible, but.. does this really matter?

bpfrancis:
In that it adds your signal samples just fine, these differences don't really matter then. @jalimena feel free to merge.

jalimena:
Ok, thanks, and sorry. We'll do better for next time.

PR #185, opened 2021-02-25
fix for including overflow and underflow in 2D histograms

jalimena:
My previous PRs on including the overflow and underflow for 2D histograms didn't account for all 8 overflow/underflow bins (previously only included 6), and some of them were sorta misplaced anyways.


jalimena:
don't merge yet. i'm going to change it such that the overflow/underflow is put into the 2D histograms in the first/last bins that frame the whole histogram, rather than just at 8 points around the border

jalimena:
Ok, should be good to review now.

bpfrancis (approved):
Changes look reasonable to me, I'll approve. I'd encourage @btcardwell to double-check the details, or perhaps double-checking the numbers in a 2D histogram work out. I'll leave it to him to merge with those then.

btcardwell:
This looks good to me and the numbers add up in every one of the handful of examples I looked through. I think this is is ready to merge.

While testing, I noticed that makePlots crashes when you use --2D without -f. I believe the lines at [0] are to blame. @jalimena, would you want to throw a fix for that into this PR as well? If not, we can always deal with it separately.

[0] https://github.com/OSU-CMS/OSUT3Analysis/blob/46b66ef4eb06b0303b2c6075c9f0b6082ef4fc80/Configuration/scripts/makePlots.py#L1298-L1304

jalimena:
good catch, thanks, @btcardwell ! i have removed the offending lines of code cuz a makeRatioPlots option doesn't make sense when we have a 2d plot anyways.

jalimena:
ok after chatting with @btcardwell , simply removing the if makeRatioPlots requirement, but keeping the position definitions is safer. I've pushed this commit, so i think we might be good to go now

btcardwell:
thanks @jalimena! I'll merge

PR #186, opened 2021-02-28
add functions for muons and electrons to get layer # of first valid pixel hit

jalimena:
from functions that determine if each layer has a valid pixel hit

### DIFF
--- a/Collections/interface/Electron.h
+++ b/Collections/interface/Electron.h
@@ -39,7 +39,17 @@ namespace osu
         const int missingMiddleHitsFromTrackerLayersWithoutMeasurements () const;
         const int missingOuterHitsFromTrackerLayersWithoutMeasurements () const;
 
-	const bool hasValidHitInPixelLayer(const uint16_t layer) const;
+	const bool hasValidHitInPixelBarrelLayer(const uint16_t layer) const;
+	const bool hasValidHitInPixelBarrelLayer1() const { return hasValidHitInPixelBarrelLayer(1); };
+	const bool hasValidHitInPixelBarrelLayer2() const { return hasValidHitInPixelBarrelLayer(2); };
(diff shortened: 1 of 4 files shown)


jalimena:
this PR has been tested and is ready for review

bpfrancis:
Looking at TrackBase.h, I'd like two changes:

1. Split up barrel and endcap functions, like here: https://github.com/OSU-CMS/OSUT3Analysis/blob/master/Collections/src/TrackBase.cc#L657 . For example you could have a muon with valid hits in PXB 1+2 and PXF 1+2+3, and be confused when there's no hit in "4" but zero "missing hits". Separate barrel/endcap methods will help.
2. Provide argument-less functions to call these, like here: https://github.com/OSU-CMS/OSUT3Analysis/blob/master/Collections/interface/TrackBase.h#L114 . We can't pass arguments in Python cuts/variable strings, and someone will eventually want these.

jalimena:
hi @bpfrancis , sure I can do this. but do you think there's something wrong with my implementation so far? based on what I have here, I made a few plots in https://milliqanelog.asc.ohio-state.edu:8080/DisplacedSUSY/1909. If you compare attachments 17 and 20, it looks like there are about 13 electrons in this sample with 0 valid pixel hits, but all electrons have their first pixel hit at layer 1? That seems incompatible, right, and so I must have done something wrong in this PR?

jalimena:
Ok, turns out I was doing something wrong in my original implementation: I would have needed the pixelHitFilter, duh! But now it is fixed, with splitting it into barrel and endcap functions and providing the argument-less functions. So the bug is fixed and everything @bpfrancis asked for is now implemented (and actually tested properly this time).

bpfrancis (approved):
Good spot on the pixelHitFilter! This looks nice to me, thanks for the changes.

btcardwell:
Looks good! I'll merge

PR #188, opened 2021-04-23
configurationOptions updates for displaced leptons (mostly new signal)

jalimena:
This PR adds a bit to configurationOptions for the displaced leptons search:
- it adds a few stop cloud model samples and composite data datasets
- it updates the HToSS masses and lifetimes, based on what was regenerated by the IFCA group
    - Here, I have left the original masses and lifetimes there but commented out on purpose. In case they decide to regenerate any more of the points they originally intended to have, we can quickly update on our end.
- it adds the new GMSB slepton samples

I've tested this with the checkLifetimeReweighting script, and also tested that I can run over these samples (at least the ones that are done so far) in 10_2_12.

### DIFF
--- a/Configuration/python/configurationOptions.py
+++ b/Configuration/python/configurationOptions.py
@@ -113,6 +113,7 @@ datasets = [
    'EGamma_2018B',
    'EGamma_2018C',
    'EGamma_2018D',
+   'DoubleEG_2016_2017_2018',
    'DoubleMu_2015',
    'DoubleMu_2015D',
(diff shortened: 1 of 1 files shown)


btcardwell:
I too have successfully run over a few of the slepton samples and would be happy merging whenever @bpfrancis is happy.

jalimena:
I've done more checks on the lifetime reweighting, all is good. I feel confident we can merge this if @bpfrancis is happy.

bpfrancis (approved):

PR #190, opened 2021-05-17
stau sample label fix

jalimena:
A minor cosmetic fix for plot legends only: GMSB stau sample labels should refer to staus and taus, not stops and tops. For displaced leptons search only


carriganm95 (approved):

jalimena:
how would you all like the review to work these days, with @bpfrancis leaving and presumably someone else taking his place as T3 manager? Do you want all of the people asked to review to sign off before merging, in general? But then in this case, this is a very straightforward PR. Can we merge already for this one?

btcardwell (approved):

btcardwell:
I have no strong opinions on the @jalimena 's general question and will happily defer to whatever @bpfrancis thinks. In this particular case, I think we can merge any time

bpfrancis:
@jalimena (and all) -- I never really thought of myself as a "manager", and thought people were requesting my review when they felt I knew something about the code they were changing. I tagged @carriganm95 and @czkaiweb as reviewers and meant to tell them later to expect that more in the future.

So for the future, I would suggest people flag reviewers they think will be knowledgable, or have some opinion. This particular PR is very uncontroversial so the exact-who wasn't vital.

jalimena:
defacto manager :)

alright, sounds good! i'll merge this one now

PR #192, opened 2021-06-04
Higgs to LL scalar samples: add a scalar mass for 125 GeV Higgs

jalimena:
adds 50 GeV scalar masses to H(125) case. for displaced leptons analysis.

tested in 94X

should be pretty straightforward


carriganm95 (approved):
looks good

PR #193, opened 2021-06-04
Update lifetime reweighting rules for sleptons samples

btcardwell:
A small update to the lifetime reweighting rules for the GMSB interpretation of the Displaced Leptons analysis. Sorry to have all these recent PRs that are just small changes to configurationOptions.

### DIFF
--- a/Configuration/python/configurationOptions.py
+++ b/Configuration/python/configurationOptions.py
@@ -9274,7 +9274,7 @@ for index, sample in enumerate(signal_datasets):
     sourceCTau = round(0.1 * 10**(math.ceil(math.log10(float(lifetime(sample))))), 5)
     # special case
     if float(lifetime(sample)) <= 0.01: sourceCTau = 0.01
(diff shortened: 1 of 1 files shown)


jalimena:
hey @btcardwell , so this PR reweights the 1000mm samples up to 10,000mm, is that right? You've checked that this is kosher for the number of events we have?

btcardwell:
Yes, this will reweight the 1,000mm samples up to 10,000mm. We don't have the stats to do this reliably yet, but we should have them soon.

Also, I imagine onfigurationOptions could evolve a bit more as we complete these final samples -- maybe we could keep this PR open for a bit and use it as a catch-all "configurationOptions updates" PR?

jalimena:
yeah sounds good

jalimena:
we are preparing a new PR that adds additional lifetime things on top of this PR, so merging this one now

jalimena (approved):

PR #194, opened 2021-06-29
fix for lifetime weights: more than one possible LLP pdgid

jalimena:
This PR implements a fix for the lifetime weights when there's more than one possible LLP pdgid, as is the case for the GMSB model for the displaced leptons search. 

For this GMSB model, you can have a LLP that is a slectron (pdg id 1000011 or 2000011), smuon (pdg id 1000013 or 2000013), or stau (pdg id 1000015 or 2000015), but only 1 abs pdgid appears in any given event. Originally, a set of lifetime weights could be assigned for each pdgid, but then if an event has another pdgid, then the weights are all set to 1, which creates at spike at 1 in the lifetime weight distribution.

Therefore, in this PR, I have made an option for "multiple pdgids", where you have only 1 set of lifetime weights (assigned to a dummy pdgid), and thus removes the spike at 1 in the individual distributions. To turn on this option, you set the bool `moreThanOneLLPType` to true, put in a number (string) for your dummy pdgid `dummyPdgId` (which you should then use in configurationOptions), and give the list of true pdgids to check (`multiplePdgIds`).


jalimena:
@btcardwell , this PR is good to go, right? can you confirm, and then maybe @carriganm95  or @czkaiweb  can take a look and merge?

btcardwell (approved):
Yes, I approve and agree @czkaiweb or @carriganm95 should merge if they approve as well.

PR #195, opened 2021-07-12
reweight HToSS125_30 samples with lifetimes>1000 mm based on lifetime=1000 mm sample

jalimena:
should only affect displaced leptons analysis


jalimena:
this PR is tested and with our code we eventually get the limit plot we now expect with more lifetimes. I think this PR is ready to merge, once others check

jalimena:
hi, can we merge?

btcardwell (approved):
Yes please on the merge. Perhaps @carriganm95 or @czkaiweb can do so if they're happy?

PR #197, opened 2021-09-14
update OSUT3Analysis to python3 and 12_1_X

jalimena:
STILL IN PROGRESS, DON'T TRY TO MERGE YET

When updating to CMSSW_12_1_X, there was an update to move to python3 (from python2) as well. The main changes that need to happen for python3 are:
- print statement syntax moved to `print()` rather than simply `print `
- `is` and `is not` sometimes give warnings that you should change them to `==` and `!=`, respectively
- integers can't start with a 0, so you have to do e.g. `0o644` when you mean the `0644` for chmod
- instead of `dictionary.has_key('keyname')` we need to use `'keyname' in dictionary`
- you have to explicitly ask for dictionary keys, values, or items as a list (https://stackoverflow.com/questions/16819222/how-to-return-dictionary-keys-as-a-list-in-python)

The ROOT version was also updated, such that ROOT Doubles are no longer a thing. I changed them all to python doubles.

All of these changes have been implemented in this PR.

### DIFF
--- a/AnaTools/scripts/checkJobs
+++ b/AnaTools/scripts/checkJobs
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import os
 import subprocess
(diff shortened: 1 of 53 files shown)


jalimena:
ok, i think these changes don't backport to earlier CMSSW releases because python2 is still the default there, so we should only merge this PR when we switch to CMSSW releases with python3

PR #213, opened 2023-01-25
Update getSiblings

carriganm95:
Update of getSiblings script to allow for usage with locally produced datasets. These local datasets have no parent/children info so run and lumi block number are used to match siblings instead. The option -l is used to indicate the dataset is local. There is also a new file getLumiBlock that uses the cmssw Lumi class to get the lumi blocks of files. This is derived from the edm tool edmLumisInFile. This class does not close root files after opening them which can result in a too many files open error (1024 on T3). The file open quota should not be changed on the T3 because we will hit a memory usage limit instead and begin using too much swap memory. Instead the subprocess package is used to get the lumi blocks of 100 files at a time and upon the end of the process the files are closed.

### DIFF
--- a/DBTools/scripts/getLumiBlock.py
+++ b/DBTools/scripts/getLumiBlock.py
@@ -1,0 +1,48 @@
+from __future__ import print_function
+from FWCore.PythonUtilities.LumiList   import LumiList
+from DataFormats.FWLite import Lumis, Handle
+import optparse
+import json
+import sys
+import os
+
+filenames = [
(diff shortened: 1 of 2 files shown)


borzari (approved):
Lines 9 - 11 from `DBTools/scripts/getLumiBlock.py` should be commented, just to avoid further confusion since they are not used. Or at least a comment stating that those were useful for debugging/testing this script is also nice. This doesn't impacts merging this PR though.

PR #220, opened 2023-04-25
Ecalo cut updates

carriganm95:
Add Ecalo and track iso from CandidateTrackProducer into DisappearingTrack class. OSUGenericTrackProducer now fills these quantities if AOD_SKIM variable is defined as true. This is set in the DisappTracks/StandardAnalysis/interface/CustomDataFormat.h file ([link](https://github.com/OSU-CMS/DisappTrks/blob/run3_devel/StandardAnalysis/interface/CustomDataFormat.h#L21)).


mlj5j (approved):
Looks good to me.  I don't see any obvious issues.

borzari (approved):
Everything looks fine

PR #222, opened 2023-05-25
Lumi mask

carriganm95:
Updates to osusub_cfg to create a lumi mask between AOD skim, MINIAOD, and golden json file. 

osusub now adds lines to the config_cfg file of jobs to create the lumi mask and apply it.

updated dataset names in configurationOptions.py to match 2022 naming scheme (JetMet, Muon, etc)


borzari (commented):
Are [these](https://github.com/OSU-CMS/OSUT3Analysis/blob/f1b1b0b2babbcecbb631d7212f288f5146180466/DBTools/scripts/getSiblings.py#L136-L150) `print`s going to stay in the code?

Where is [this](https://github.com/OSU-CMS/OSUT3Analysis/blob/f1b1b0b2babbcecbb631d7212f288f5146180466/DBTools/scripts/getSiblings.py#L173) function defined?

PR #224, opened 2023-07-26
getSiblings Updates

carriganm95:
Updates to the getSiblings file in DBTools/scripts...

1. Make getSiblings into a class
2. Option (--lumiMatching) to force lumi matching instead of DBS parent/child
3. Lumi lists are now created by DBS api rathern than FWCore Lumi class
4. Options for prod type of datasets: see below
5. Argument to pass input/sibling json files of {filename: [lumis]} to avoid generating them

Prod Types:

- global: both datasets are global in DAS
- user: input dataset is user created (phys03) and sibling is global
- allUser: both datasets are user created
- local: input dataset is local to T3 (lumi matching automatic)
- allLocal: both datasets are local to T3 (lumi matching automatic)


borzari (approved):
Just tested it and worked fine for one of the use cases. Will test other scenarios as they come up, but this can be merged already.

Issue #29, opened 2016-01-14
makePlots.py "-e -s" options incompatible


lantone:
it fails when both options are used at once

jalimena:
this one was never fixed, i imagine, but it's not a typical use case. Worth trying to fix?

PR #226, opened 2023-08-21
Update types

carriganm95:
Update to be able to run on skim files. Merge out no longer makes a pkl file and instead produces a json file for the skim collection types.

Edits to OSUGenericTrackProducer to run on isolatedTracks instead of generalTracks collections. 

Updates to getSiblings.py script. There are now options to do lumi matching instead of DAS sibling matching. Can also run on global, user, local files.

Update configuration options dataset names following updates in DisappTrks PR #29.


mlj5j (approved):
I don't see any issues.

mlj5j (approved):
Only thing I might change is updating the MET_2022E-G dataset names to JetMET in configurationOptions.py, but another PR can take care of that.

PR #230, opened 2023-11-24
Event Mask

carriganm95:
Generate and apply an event mask when running osusub.

Uses getSiblings to create a list of AOD/miniAOD file siblings
Creates list of events shared by these files and applies the event list as an event mask on the job

List is created as a step prior to running cmsRun

Note: currently configured to always create this event masking. Should only be applied during first step (no cuts step).


DBTools/python/getSiblings.py
```diff
@@ -0,0 +1,331 @@
...
+    def getFilesFromList(self, jobNumber, nJobs):
+        sys.path.append(os.getcwd())
+
+        Label = args.inputFiles
+        datasetInfo = importlib.import_module('datasetInfo_' + Label +'_cfg', package=None)
+
+        #exec("import datasetInfo_" + Label +"_cfg as datasetInfo")
+
+        filesPerJob = int(1) #int (math.floor (len (datasetInfo.listOfFiles) / nJobs))
+        residualLength = int(10) #int(len(datasetInfo.listOfFiles)%nJobs)
```
borzari, line 285: Is line 282 really needed?

Is there a reason to keep comments in lines 284 and 285?

DBTools/python/getSiblings.py
```diff
@@ -0,0 +1,331 @@
...
+            self.getDBSSiblings(self.dictName)
+
+    #function to get siblings from DBS
+    def getDBSSiblings(self, output_json):
+        print("getting dbs siblings")
+
+        file_dict = {}
+
+        print("Output JSON file will be called {0}".format(self.dictName))
+        #print('Input dataset: {}'.format(self.dataset_in))
```
borzari, line 126: Maybe remove this line or add a debug/verbose option to print it if it can be useful for debugging

DBTools/interface/getEvents.h
borzari, line 1: Wouldn't it be easier/better to have a `runInfo` struct and a function `setRunInfo` that assigns the values of `runNum`, `lumiBlock` and `event`, instead of putting `runInfo` inside itself?

DBTools/python/osusub_cfg.py
```diff
@@ -141,10 +151,10 @@ def getRun3SkimSiblings (fileName, dataset, inputUser='global'):
           grandparents.extend (dbs3api_phys03.listFileParents(logical_file_name = parent_file_name))
       parents = grandparents
     elif inputUser == 'user':
-      print("Getting user dataset, not grandparents")
+      #print("Getting user dataset, not grandparents")
       parents = dbs3api_phys03.listFileParents (logical_file_name = fileName)
     else:
-      print("Getting user dataset, not grandparents")
+      #print("Getting user dataset, not grandparents")
```
borzari, line 157: Are these comments needed?

borzari (commented):
I put some comments in specific files, and have two other general ones:

- What is the need for all of the `json*` files?
- Did you tested the several distinct use cases? For data/MC that is local/remote?

DBTools/interface/getEvents.h
carriganm95: I am a bit confused with what you are suggesting. I only use the struct when calling it with a constructor that sets the values for run, lumi, and event

carriganm95:
> I put some comments in specific files, and have two other general ones:
> 
> * What is the need for all of the `json*` files?
> * Did you tested the several distinct use cases? For data/MC that is local/remote?

These were part of a previous attempt to create the lumi mask using json files. I have removed.

I have only tested with data because I know that was the pressing issue. I would suggest we at least switch to this branch for the time being and can test other cases. This way we do not hold up data processing.

DBTools/python/osusub_cfg.py
carriganm95: fixed

DBTools/python/getSiblings.py
carriganm95: removed

DBTools/python/getSiblings.py
carriganm95: these were supposed to be changed back, fixed now

borzari:
> These were part of a previous attempt to create the lumi mask using json files. I have removed.
> 
> I have only tested with data because I know that was the pressing issue. I would suggest we at least switch to this branch for the time being and can test other cases. This way we do not hold up data processing.

You are completely right. I think it is possible to use this branch for data while testing for the other cases. Not even sure if it will be needed in other cases as for signal MC the lumi matching works fine and for BG MC we have the MiniAOD + skim procedure.

DBTools/interface/getEvents.h
borzari: Ah, I got confused, sorry. That is completely fine.

carriganm95:
Options added to only merge and create event mask with command line argument "--mergeSkim"

DBTools/python/osusub_cfg.py
```diff
@@ -186,13 +207,18 @@ def getSiblingList(sibList, runList, siblingDataset):
   fin = open(sibList, 'r')
   data = json.load(fin)
 
+  aux_int = 0
```
borzari, line 210: I added this "counter" to keep track of number of files when the number of events is not -1, but I am not sure if it was the best way to go. Anyone has any comments/suggestions on how to change it, or to leave it as is?

DBTools/scripts/getSiblings.py
borzari, line 1: Is this file really needed since we have the one in `DBTools/python`?

carriganm95: removed

DBTools/python/osusub_cfg.py
carriganm95: I think this is fine, my only other suggestions would be `for filename in runlist[:len(data)]` basically the same thing but no counter

borzari: @carriganm95 I am not really sure this would work because sometimes runList includes files that don't contain siblings and should be skipped (I am seeing this in the `allLocal` case), and then the number of siblings doesn't take those skips into account. I am doing some more tests to check it, and maybe will commit another change to this part of the config

borzari (approved):

PR #236, opened 2024-04-23
Get siblings fix

carriganm95:
Fixes get siblings by reducing calls to databases

replaces dbs calls with shell DAS calls to get file run and lumi block info

lists of primary/sibling file run/lumi info are saved to json files

at run time individual jobs have MINIAOD/AOD matched based on run/lumi block info
at this time an event list is also created to get only events shared between the two

This has been tested using:
data: EGamma_2022F
background MC: /QCD_PT-2400to3200_TuneCP5_13p6TeV_pythia8/Run3Summer22MiniAODv3-124X_mcRun3_2022_realistic_v12-v2/MINIAODSIM
signal MC (local): /data/users/borzari/condor/SignalMC/Run3/2022/step3/100cm/100GeV/AMSB_chargino_M_100GeV_CTau_100cm_TuneCP5_PSweights_13p6TeV_madgraph5_pythia8/

### DIFF
--- a/DBTools/python/getSiblings.py
+++ b/DBTools/python/getSiblings.py
@@ -28,7 +28,7 @@ r.gSystem.Load('libOSUT3AnalysisDBTools.so')
 # class used to get the secondary files (siblings) of the primary input files
 class getSiblings():
 
-    def __init__(self):
+    def __init__(self, inputJSON, secondaryJSON, label):
 
         self.numEvents = 0
         self.numSharedEvents = 0
@@ -38,7 +38,15 @@ class getSiblings():
         self.dataset_in = None
         self.dataset_sib = None
 
-        if not args.eventList:
+        self.inputJSON = inputJSON
+        self.secondaryJSON = secondaryJSON
+        self.label = label
+
+        self.nJobs = -1
+        self.jobNumber = -1
+        self.eventsPerJob = -1
+
+        '''     if not args.eventList:
 
             #if all datasets are not local set APIs
             if args.prod != 'allLocal':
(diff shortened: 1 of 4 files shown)


DBTools/python/getSiblings.py
```diff
@@ -113,13 +131,193 @@ def setAPIs(self):
...
+            self.getFilesFromList(args.jobNumber, args.totalJobs)
+            self.findMatches()
         else:
-            self.getDBSSiblings(self.dictName)
+            self.getFilesFromList(args.jobNumber, args.totalJobs)
+            self.findMatches()
+
+    @staticmethod
+    def getDASInfo(dataset, jsonName=None):
+        cmd = 'dasgoclient -query="file,run,lumi,events dataset={0}" -json'.format(dataset)
```
borzari, line 150: Are the json files large when also asking for events? If so, could it be an issue with storage space?

mlj5j:
Can’t be that big, right?  I guess Mike could comment on the actual size, but I think they are probably not big enough to be a problem.

Matt

From: borzari ***@***.***>
Date: Tuesday, April 23, 2024 at 3:48 PM
To: OSU-CMS/OSUT3Analysis ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [OSU-CMS/OSUT3Analysis] Get siblings fix (PR #236)
@borzari commented on this pull request. In DBTools/python/getSiblings. py: > else: - self. getDBSSiblings(self. dictName) + self. getFilesFromList(args. jobNumber, args. totalJobs) + self. findMatches() + + @staticmethod + def getDASInfo(dataset,


@borzari commented on this pull request.

________________________________

In DBTools/python/getSiblings.py<https://urldefense.com/v3/__https:/github.com/OSU-CMS/OSUT3Analysis/pull/236*discussion_r1576297726__;Iw!!KGKeukY!yeHH7qO6QlG38fNNavoRecauCQu_xT9sw75cJWeuyrrOY2iS6ns2-q0dc3KFhKoUmYMfjCOVTd0VM7GBxToD8Ikmzz-Y$>:

>          else:

-            self.getDBSSiblings(self.dictName)

+            self.getFilesFromList(args.jobNumber, args.totalJobs)

+            self.findMatches()

+

+    @staticmethod

+    def getDASInfo(dataset, jsonName=None):

+        cmd = 'dasgoclient -query="file,run,lumi,events dataset={0}" -json'.format(dataset)

Are the json files large when also asking for events? If so, could it be an issue with storage space?

DBTools/python/getSiblings.py
```diff
@@ -113,13 +131,193 @@ def setAPIs(self):
...
+
+        miniDict = {}
+
+        for filename in os.listdir(dataset):
+            if not filename.endswith('.root'): continue
+            print("Working on file", dataset+filename)
+            lumi_list = []
+            run_list = []
+
+            fin = r.TFile.Open(dataset + filename, 'read')
```
borzari, line 183: Given the issue we were seeing with the local files, and since this is a "new" way of accessing it, this might be the source of illegal instructions. Not sure why or how, but it is the only distinction from the previous getSiblings. I would suggest to use the `r.getEventsInFile()` here to get the lumis (also to avoid the memory issues we saw a while ago, which are not very relevant for the local case, but still)

borzari (commented):
I made a few comments during the review. Will mark as approved after discussing about those.

DBTools/python/getSiblings.py
carriganm95: largest I have seen was EGamma_2022F for miniAOD at 1.5Mb so shouldn't be an issue

borzari (approved):

PR #242, opened 2024-05-15
Update cmssw 13

carriganm95:


borzari (commented):
Are all files in `Anatools/data/scaleFactors/electron2022` really needed?
Everything that is added to the `collections/plugins/OSUGenericTrackProducer.cc` file is for the ML approaches? The electron, jet and tau IDs in that file need to be updated for Run 3 (if I am not mistaken). Also, the `is2017_` flag might not be needed there.
I'll check the HLT object matching to see if everything is reasonable. The names of the objects/filters in the HLT may be misleading sometimes.
The rest, if it is all validated, seems to be fine

borzari (approved):

PR #250, opened 2024-08-21
updates to run with crab

carriganm95:


borzari (approved):
Small question: the .pb files just changed place, or are those new files?

carriganm95:
> Small question: the .pb files just changed place, or are those new files?

you are right, it was changes to make sure files were transferred with crab. I have consolidated the locations

borzari (approved):

PR #257, opened 2024-10-15
Changing jet res and electron d0, dz cuts for run 3

borzari:
This PR sets `jetResNewPrescription` as true also for Run 3 jets to use the recommendation from the JME POG.

It also changes the electron cut based d0 and dz selections to use the recommendation from the EGM POG.


Configuration/python/CollectionProducer_cff.py
```diff
@@ -312,6 +313,14 @@ class CollectionProducer:
     EERecHits          =  cms.InputTag  ("reducedEcalRecHitsEE"),
```
carriganm95, line 313: just to make sure I understand, you want to leave the AOD rec hits as default and only use miniAOD selectively?

borzari: Exactly. I'll try doing something more permanent/automatic, but will leave it for a future PR if we actually go with the MiniAOD-only approach (after checking the deepsets case). I added that here now to not forget this approach, which is very simple to avoid using AOD somewhere.

borzari: Talked in private with @carriganm95 and decided that this PR can be merged.

PR #271, opened 2025-02-12
Fiducial Map2023D

carriganm95:
fiducial maps for 2023D

remove option where era-by-era fiducial maps are automatically set in osusub


Configuration/python/CollectionProducer_cff.py
```diff
@@ -401,13 +401,6 @@ class CollectionProducer:
         collectionProducer.tracks.fiducialMaps.muons[0].histFile = cms.FileInPath ("OSUT3Analysis/Configuration/data/muonFiducialMap_2022F_data.root")
     else:
         print("No fiducial map hist file")
-    # determine which era this dataset is in
-    if "_201" in osusub.datasetLabel:
-        collectionProducer.tracks.fiducialMaps.electrons[0].era = cms.string (osusub.datasetLabel[osusub.datasetLabel.find('_201'):])
-        collectionProducer.tracks.fiducialMaps.muons[0].era = cms.string (osusub.datasetLabel[osusub.datasetLabel.find('_201'):])
```
borzari, line 407: I would say that these can be left in the code. It is only for Run 3 (202X) that we have the "era-by-era" fiducial maps already divided into separate root files.

borzari (approved):

Configuration/python/CollectionProducer_cff.py
carriganm95: ok done

PR #290, opened 2026-01-13
Use trimmed JSON file for JERC corrections

lnestor:
Setting as draft because I want to add in verification that this doesn't change anything. Also I may trim the JSON file even more.


lnestor:
This did not improve memory usage.

PR #294, opened 2026-02-05
Fix ROOT warning by removing duplicate in classes_def.xml

lnestor:
Every time we import ROOT or use ROOT in anyway, I was always getting these warnings which were annoying. This fixes those warnings. I'm not merging directly to master because I need to test if it messes things up when jobs are run.


lnestor:
This prevented CMSSW jobs from running, closing

